PR #6, opened 2014-09-03
8.0 port account analytic required

adrienpeiffer:
Port account analytic required to 8.0


account_analytic_required/__openerp__.py
```diff
@@ -45,6 +45,6 @@
     'website': 'http://www.akretion.com/',
     'depends': ['account'],
     'data': ['account_view.xml'],
-    'installable': False,
-    'active': False,
+    'installable': True,
+    'active': True,
```
pedrobaeza: Please don't put active as True. This will install the module as soon as you create the DB. In fact, active is deprecated in favour of 'auto_install' key.

pedrobaeza:
Due to the migration, can we take advantage of this to make a test that also checks if migration if correct?

sbidoul:
@pedrobaeza there were already tests for account_analytic_required in the 7.0 branch (I wrote them myself a couple of month ago). In this case I see the tests succeeding in travis. Otherwise travis fails here due to PEP8 errors in unported modules.

So :+1: from my side.

That said, I don't think we must mandate the presence of tests for a mere port to 8.0. We have a lot of porting to do and we cannot affort to write all missing tests at once. I suggest tests are made mandatory for new features and/or significant changes to existing modules.

gurneyalex:
@sbidoul flake8 should not be reporting errors on addons in **unported**, if it does, then the travis config is wrong :unamused:

sbidoul:
Indeed the travis config is quite old for account-analytic. I'll do a PR to bring it up-to-date.

gurneyalex:
On 08/09/2014 09:59, Stéphane Bidoul (ACSONE) wrote:

> Indeed the travis config is quite old for account-analytic. I'll do a
> PR to bring it up-to-date.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/account-analytic/pull/6#issuecomment-54786784.
> 
> I'm on it.

## 

Alexandre Fayolle
Chef de Projet
Tel : + 33 (0)4 79 26 57 94

Camptocamp France SAS
Savoie Technolac, BP 352
73377 Le Bourget du Lac Cedex
http://www.camptocamp.com

pedrobaeza:
@sbidoul, sorry, I didn't see the test. I know there's a lot of work making the migration, but it's a good oportunity to have the test doing the work of checking if the migration is correct.

:+1: for this PR, and waiting the PR for the Travis config.

Regards.

gurneyalex:
@sbidoul that's #7

gurneyalex:
@sbidoul I would not mandate tests for code ported to v8 when the port is straightforward (but I will gently nudge the porter into providing these). OTOH, when the porter does the work of porting the addon to the new API, then I think tests must be provided.

sbidoul:
Rebased, travis green :v: 

@gurneyalex agreed.

alexis-via:
OK, let's merge this then ! I hate these projects with empty branches for v8 :)

@adrienpeiffer auto_install defaults to False, so I don't see a good reason to add the key in **openerp**.py.

foutoucour:
:+1:

PR #15, opened 2014-09-09
rewrite store trigger: override breaks

### LINKED ISSUE #14: crash when writing the company when analytic_multicurrency is installed
To reproduce on v7:
- install `analytic_multicurrency`
- write any field in the company

You get the following crash: 

```
TypeError: super(type, obj): obj must be an instance or subtype of type
```

The traceback points to the method causing the problem: https://github.com/OCA/account-analytic/blob/7.0/analytic_multicurrency/analytic.py#L100-L102.

That breaks because a store trigger in OpenERP is a method of a class which is called as a method of another class. In that case, we are in the `account_analytic_account` class, but `self` is a `res_company`.

lepistone:
Fixes #14. Store triggers cannot be overridden because of
https://bugs.launchpad.net/openobject-server/+bug/893079

### DIFF
--- a/analytic_multicurrency/analytic.py
+++ b/analytic_multicurrency/analytic.py
@@ -98,8 +98,20 @@ class account_analytic_account(orm.Model):
         return result
 
     def _get_analytic_account(self, cr, uid, ids, context=None):
-        return super(account_analytic_account, self)._get_analytic_account(
-            cr, uid, ids, context=context)
+        """Copied from the original in the core.
+
+        Store triggers cannot be overridden because of
+        https://bugs.launchpad.net/openobject-server/+bug/893079
+
+        """
+        company_obj = self.pool.get('res.company')
+        analytic_obj = self.pool.get('account.analytic.account')
+        accounts = []
+        for company in company_obj.browse(cr, uid, ids, context=context):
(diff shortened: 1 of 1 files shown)


vrenaville:
LGTM :+1:

analytic_multicurrency/analytic.py
```diff
@@ -98,8 +98,20 @@ def _currency(self, cr, uid, ids, field_name, arg, context=None):
...
+        Store triggers cannot be overridden because of
+        https://bugs.launchpad.net/openobject-server/+bug/893079
+
+        """
+        company_obj = self.pool.get('res.company')
+        analytic_obj = self.pool.get('account.analytic.account')
+        accounts = []
+        for company in company_obj.browse(cr, uid, ids, context=context):
+            accounts += analytic_obj.search(cr, uid, [
+                ('company_id', '=', company.id)
```
PCatinean, line 112: In this context isn't accounts = analytic_obj.search(cr, uid, [('company_id','in',ids)]) shorter?

lepistone, line 112: I think you're right @PCatinean, but IMO it is easier to keep the same code as the original.

tafaRU:
:+1:

PR #17, opened 2014-10-29
Forward port branch '7.0' until 421a016

sbidoul:
I did a merge of 7.0 in 8.0 and resolved the minor conflicts.


pedrobaeza:
Thanks for taking the time to make the cherry-pick merge.

:+1:

pedrobaeza:
Please review this soon for not getting old and useless.

account_analytic_plan_required/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from . import account
```
yvaucher, line 23: Shouldn't this module account_analytic_plan_required go in `__unported__` ?

sbidoul, line 23: @yvaucher account_analytic_plan_required works as is on 8.0 too, so no need to go to unported.

yvaucher:
Ok then :+1:

sbidoul:
@yvaucher can you remove the needs fixing tag?

clonedagain:
Why are there commits from 2013?

clonedagain:
Is there a magic git command to show only the parts where a conflict was resolved? It would make the review much more feasible.

sbidoul:
@clonedagain if you follow the lines in gitk, you see that these commits from 2013 were merged in 7.0 after 8.0 was branched. Hence they show up here when merging 7.0 in 8.0.

I don't know the magic git command you are looking for. I you look at commit ffd55a1 in gitk, it shows the diff of the conflict resolution only.

I agree these kind of diffs are not easy to review. I believe we should do these forward ports more often, otherwise we'll end up rediscovering bug fixes in both branches. The new API is not helping, though, and I know in some repos it's already become virtually impossible to forward port automatically, but at least we should do the effort of merge+manual resolution so we can track what has been forward ported and what not.

pedrobaeza:
The method I have when migrating a module, is to catch the version from 7.0 repository and then start to migrate. In this way, I don't loose any patch applied to 7.0 and not 8.0.

Regards.

clonedagain:
@pedrobaeza yes but you'll still need to forward-port later bugfixes from 7.0 to upper versions.
On the ML it was agreed that most OCA projects would not be able to afford to to this kind of ports, I'm glad to see this happen.

pedrobaeza:
Yeah, of course. This can happen if 8.0 modules are not migrated to new API, but we have to do it, because old API is removed in 9.0, so this transition version is going to be difficult. Let's hope there is not many bugs to port/backport!

sbidoul:
@pedrobaeza @clonedagain if we do the git merge 7.0 in 8.0 often (which possibly means redoing the fix in 7.0 manually in 8.0), at least we let git mark what has been forward ported.

sbidoul:
So, to sum up, all commits in this PR come from the 7.0 branch.

The conflict resolution is in ffd55a1, best viewed in gitk (as github shows the whole merge diff). The rest has been automerged.

adrienpeiffer:
:+1:  code review

PR #24, opened 2015-02-19
6.1 account analytic project

NL66278:
Add possibility to create project from analytic account that was not a project before. Also provided link between analytic account and project,

### DIFF
--- a/account_analytic_project/__openerp__.py
+++ b/account_analytic_project/__openerp__.py
@@ -30,8 +30,9 @@ Withouth this module, if you first create an analytic account and later on you
 decide to have a project as well for this account, the situation is very
 difficult to rectify.
 """,
-    'version': '6.1.r001',
+    'version': '1.0',
     'author': 'Therp BV',
+    'category': 'Generic Modules/Accounting',
     'website': 'http://therp.nl',
     'email': 'info@therp.nl',
     'depends': [
@@ -42,3 +43,5 @@ difficult to rectify.
         'view/account_analytic_account.xml',
     ],
 }
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
--- a/account_analytic_project/__openerp__.py
+++ b/account_analytic_project/__openerp__.py
@@ -1,0 +1,44 @@
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2015 Therp BV <http://therp.nl>
+#    All Rights Reserved
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Project and analytic account integration',
+    'description': """\
+Project and analytic account integration.
+
+Gives the option to create projects directly from the analytic accounts they
+should be linked to.
+
+Withouth this module, if you first create an analytic account and later on you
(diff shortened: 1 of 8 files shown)


account_analytic_project/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+##############################################################################
+{
+    'name': 'Project and analytic account integration',
+    'description': """\
+Project and analytic account integration.
+
+Gives the option to create projects directly from the analytic accounts they
+should be linked to.
+
+Withouth this module, if you first create an analytic account and later on you
```
hbrunn: -h

account_analytic_project/model/account_analytic_account.py
```diff
@@ -0,0 +1,112 @@
...
+        res = {}
+        for analytic_id in ids:
+            has_project = False
+            project_ids = project_model.search(
+                cr, uid, [('analytic_account_id', '=', analytic_id)],
+                context=context
+            )
+            if project_ids:
+                has_project = True
+            res[analytic_id] = has_project
```
hbrunn: why not bool(project_ids)?

hbrunn:
some little nags, rest looks goot to me. One architectural question: Why don't you introduce the inverse to project.analytic_account_id here and use that in your checks? Then the ORM does all the searching for you, you don't need the function field and other code can simply loop through projects if necessary.

account_analytic_project/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Project and analytic account integration',
+    'description': """\
```
pedrobaeza: Remove \

account_analytic_project/i18n/account_analytic_project.pot
```diff
@@ -0,0 +1,51 @@
...
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: account_analytic_project
+#: code:addons/account_analytic_project/model/account_analytic_account.py:64
+#, python-format
+msgid "This method is only valid for a single analytic account"
+msgstr "This method is only valid for a single analytic account"
```
pedrobaeza: pot file mustn't have strings on msgstr.

account_analytic_project/model/account_analytic_account.py
pedrobaeza: And also a `break` to avoid consuming the rest of the loop

NL66278: But theoretically the method could be called for multiple rows. And if not, the loop will be done in one iteration anyway.

account_analytic_project/i18n/account_analytic_project.pot
NL66278: The pot file is the mostly unaltered result of exporting the English language from the module. That has all the msgstr's filled. Is this a problem? It is not a big problem to remove them from the few lines for this module, but other modules might have a lot of text, so If not needed, I would be glad to leave them the way they were exported.

pedrobaeza: No, you can export an empty file if you select no language in the wizard. That's the correct pot file.

NL66278: Thanks for the information. That was totally new for me.

hbrunn:
[retracted thumbs up]

account_analytic_project/model/account_analytic_account.py
```diff
@@ -0,0 +1,109 @@
...
+        return {
+            'type': 'ir.actions.act_window',
+            'name': action_name,
+            'view_mode': view_mode,
+            'view_type': 'form',
+            'view_id': view_id,
+            'res_model': 'project.project',
+            'nodestroy': True,
+            'res_id': res_id,
+            'target':'current',
```
hbrunn: ./account_analytic_project/model/account_analytic_account.py:99:21: E231 missing whitespace after ':'

hbrunn:
:+1: (flake8 and pylint errors are caused by other modules)

pedrobaeza:
:+1:

PR #26, opened 2015-03-04
8.0 update account_analytic_required

luc-demeyer:
This PR makes the analytic account a mandatory field on the invoice line when the analytic policy for the selected account is equal to 'always'.


account_analytic_required/account_invoice.py
```diff
@@ -0,0 +1,37 @@
...
+class account_invoice_line(models.Model):
+    _inherit = 'account.invoice.line'
+
+    @api.model
+    def _get_policies(self):
+        return self.pool["account.account.type"]._get_policies()
+
+    analytic_policy = fields.Selection(
+        '_get_policies', string='Policy for analytic account',
+        related='account_id.user_type.analytic_policy')
```
sbidoul: Is the compute method necessary in addition to the related?

account_analytic_required/__openerp__.py
```diff
@@ -23,28 +23,16 @@
...
-
-Module developped by Alexis de Lattre <alexis.delattre@akretion.com>
-during the Akretion-Camptocamp code sprint of June 2011.
-""",
-    'author': "Akretion,Odoo Community Association (OCA)",
-    'website': 'http://www.akretion.com/',
+    'summary': 'Account Analytic Required',
+    'author': "Akretion,"
+              "Noviat,"
+              "Odoo Community Association (OCA)",
```
sbidoul, line 31: What's the policy regarding this authors field?

I personally don't mind you add Noviat. But then you probably need to add all contributors.

sbidoul:
Overall I find this interesting, but I'm a bit worried customer who suddenly see this field becoming mandatory on invoices will ask the same behaviour everywhere else (move lines, journal entries, write-off screen, bank statement, etc).

account_analytic_required/__openerp__.py
pedrobaeza, line 31: As contributors is a field that Odoo doesn't accept, the way to put contributors is in author key, now that Odoo admits in their app store multiple authors separated by comma for the modules. About OCA inclusion, is a way to have all OCA modules tagged.

account_analytic_required/account_invoice.py
luc-demeyer: @sbidoul 
I tested without compute method and it works.
I have updated the PR. 
Thanks for the valuable suggestion.

luc-demeyer:
@sbidoul 
I agree with your remark about worried end-users but now it's worse. 
e.g. you are encoding a supplier invoice with 40 lines and only at save time you get an error message that you should have added the analytic account. 
With invoices we cover the most cases, write off is only one line hence the end-user has an easy one-to-one relationship between the  constraint error message and the write off account (and you can predefine an analytic default on this one). Miscellaneous journal entries is probably the next priority (but these are usually encoded by the accountant, not a assistant doing mass encoding of supplier invoices). 
Hence I believe this PR is a substantial improvement of the module, but it's correct that adding the same logic to more objects is a next logical step.

luc-demeyer:
I have updated the PR, the analytic field now becomes a mandatory field on the User Interface for all basic account module objects : account.invoice, account.move.line, account.bank.statement

luc-demeyer:
can someone tell me what is required to 'greenify' this PR ?

pedrobaeza:
Include tests that don't decrease the coverage. Anyway, tests are not mandatory yet, but let's try to keep the good achieved coverage

sbidoul:
@luc-demeyer can you rebase to resolve the conflict? We'll review it.

account_analytic_required/__openerp__.py
```diff
@@ -23,28 +23,18 @@
 
 {
     'name': 'Account Analytic Required',
-    'version': '0.2',
+    'version': '0.5',
```
adrienpeiffer: Bad version format. Can you rebase your branch to get the correct format ?

account_analytic_required/models/account_move_line.py
```diff
@@ -21,41 +21,62 @@
...
+        "'Optional', the accountant is free to put an analytic account "
+        "on an account move line with this type of account ; if you "
+        "select 'Always', the accountant will get an error message if "
+        "there is no analytic account ; if you select 'Never', the "
+        "accountant will get an error message if an analytic account "
+        "is present.")
+
+
+class AccountAccount(models.Model):
+    _inherit = "account.account"
```
adrienpeiffer: Already redefined in `account_analytic_required/models/account_account.py`

luc-demeyer:
rebased + obsolete code removed from account_move_line.py

account_analytic_required/models/account_move_line.py
```diff
@@ -24,47 +24,6 @@
...
-        help="Set the policy for analytic accounts : if you select "
-        "'Optional', the accountant is free to put an analytic account "
-        "on an account move line with this type of account ; if you "
-        "select 'Always', the accountant will get an error message if "
-        "there is no analytic account ; if you select 'Never', the "
-        "accountant will get an error message if an analytic account "
-        "is present.")
-
-
-class AccountAccount(models.Model):
```
adrienpeiffer: Why do you remove this class ?

adrienpeiffer:
:+1: (Code review and functional test)

account_analytic_plan_required/tests/test_account_analytic_plan_required.py
```diff
@@ -148,8 +148,11 @@ def test_never_no_analytic(self):
 
     def test_never_with_analytic(self):
         self._set_analytic_policy('never')
-        with self.assertRaises(orm.except_orm):
-            self._create_move(with_analytic=True, with_analytic_plan=False)
+        # with self.assertRaises(orm.except_orm):
+        #     self._create_move(with_analytic=True, with_analytic_plan=False)
```
yvaucher, line 152: Why is this commented ?

account_analytic_required/README.rst
```diff
@@ -0,0 +1,58 @@
...
+=======
+
+Author
+------
+* Module developed by Alexis de Lattre <alexis.delattre@akretion.com>
+  during the Akretion-Camptocamp code sprint of June 2011.
+
+Contributors
+------------
+* Luc De Meyer, Noviat <info@noviat.com>
```
yvaucher, line 44: Missing a bunch of people in credits: https://github.com/OCA/account-analytic/commits/7.0/account_analytic_required

account_analytic_required/models/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
yvaucher, line 21: In `__init__.py` files we can remove license headers when there is no code

account_analytic_required/__openerp__.py
```diff
@@ -23,28 +23,18 @@
...
-set the policy to *always* for the account type *expense* ; then, if you
-try to save an account move line with an account of type *expense*
-without analytic account, you will get an error message.
-
-Module developped by Alexis de Lattre <alexis.delattre@akretion.com>
-during the Akretion-Camptocamp code sprint of June 2011.
-""",
-    'author': "Akretion,Odoo Community Association (OCA)",
-    'website': 'http://www.akretion.com/',
+    'summary': 'Account Analytic Required',
```
yvaucher, line 29: Don't repeat the title in the summary.

account_analytic_required/static/src/js/account_analytic_required.js
```diff
@@ -0,0 +1,115 @@
...
+
+    instance.web.account.bankStatementReconciliationLine.include({
+
+        init: function(parent, context) {
+            this._super.apply(this, arguments);
+            this.map_analytic_policy = this.getParent().map_analytic_policy;
+            this.required_fields_set = this.getParent().required_fields_set;
+            },
+
+        UpdateRequiredFields: function(elt) {
```
yvaucher, line 72: Functions should be in snake case as in Odoo core modules. `update_required_fields` or at least they should start with a lower case.

account_analytic_required/static/src/js/account_analytic_required.js
```diff
@@ -0,0 +1,115 @@
...
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+*/
```
yvaucher, line 22: See https://github.com/OCA/maintainer-tools/blob/master/template/module/static/src/js/module_name.js

For the new template for license header

pedrobaeza:
We are going to close this one as this is no longer used by Noviat. If you want to continue, please feel free to take this code and improve the comments.

PR #29, opened 2015-04-14
[MIG] 8.0 migration of analytic_secondaxis

AdilHoumadi:


analytic_secondaxis/__openerp__.py
```diff
@@ -32,46 +34,47 @@
...
     "name": "Account Analytic Second Axis",
     "version": "1.1",
-    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "author": "Camptocamp,Taktik,Odoo Community Association (OCA)",
     "category": "Generic Modules/Accounting",
     "description":
-    """
-    Add a second analytical axis on analytic lines allowing you to make
-    reporting on.
+        """
```
pedrobaeza: Extract this into README.rst file and use OCA's template (https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst)

analytic_secondaxis/analytic_secondaxis.py
```diff
@@ -0,0 +1,291 @@
...
+            count=count
+        )
+
+    @api.multi
+    def _get_first_aa_which_have_activity(self, account):
+        """Return browse record list of activities of the account which
+        have an activity set (goes bottom up, child, then parent)
+        """
+        account.ensure_one()
+        if account.activity_ids:
```
pedrobaeza: Better to make it iterative instead recursive.

analytic_secondaxis/analytic_secondaxis.py
```diff
@@ -0,0 +1,291 @@
...
+import openerp.addons.decimal_precision as dp
+
+
+class ProjectActivityAl(models.Model):
+    _name = "project.activity_al"
+    _description = "Second Analytical Axes"
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        analytic_account_ids = []
```
pedrobaeza: Maybe it's better to have a display_name and change rec_name to that one than performing this complicated searches.

analytic_secondaxis/analytic_secondaxis.py
```diff
@@ -0,0 +1,291 @@
...
+                )
+            else:
+                return False
+
+    @api.model
+    def name_search(self, name, args=None, operator='ilike', limit=80):
+        """Override name_search function that perform the search
+        on the name of the activities
+        """
+        if not args:
```
pedrobaeza: Maybe it's better to have a display_name and change rec_name to that one than performing this complicated searches.

analytic_secondaxis/__openerp__.py
AdilHoumadi: Done in 9ef20c28dcdc1f8296e27eba88f9938c93d14a40.

pedrobaeza: You have to remove it from here then.

analytic_secondaxis/models/analytic_secondaxis.py
```diff
@@ -0,0 +1,263 @@
...
+    )
+    name = fields.Char(
+        string="Activity",
+        required=True,
+        size=64,
+        translate=True
+    )
+    display_name = fields.Char(
+        string='Display Name ',
+        compute=_compute_display_name,
```
pedrobaeza: Put method name between quotes to make it easily inheritable.

pedrobaeza:
You don't need to change all double quotas to single. They act the same.

As you can see, default methods cannot be enclosed by quotes, because it can't be distinguished between an string value and a method.

AdilHoumadi:
Thanks for you explanation, It's clear now, I'll take it in consideration.
Thank you for this review !

PR #35, opened 2015-08-12
[ADD] analytic_partner

pedrobaeza:
# Classify analytic items by partner

This module adds a partner on each analytic item for allowing to have another
dimension for analysing data.

It also handles the proper propagation of this field to the created analytic
entries when validating invoices.
# Configuration

You have to be granted as at least "Accountant" in your user profile and
have checked the "Analytic Accounting" access right.
# Usage

Go to Accounting > Analytic Journal Items, and there, you can set the partner
for the analytic items, and search or group by it.

You can also go to a partner, and click on the smart-button "Cost/Revenue",
placed on the upper-right part, and you will navigate to the analytic items
associated to this partner.
# Known issues / Roadmap
- This module hasn't been tested with _account_analytic_plans_ module
  installed, so maybe it's incompatible with it.


rafaelbn:
- To test it in runbot first you have to go: settings - users - admin, and set "Accounting & Finance" as Financial Manager. If not you cannot see this:

> Go to Accounting > Analytic Journal Items, and there, you can set the partner
> for the analytic items, and search or group by it.
- In the form view of account.analytic.line the Partner's fields is below "analytic account" if you place it bellow "Ref." it will be in the same position as Odoo v9. So after migration the fields will be in the same place.

Functional test :+1:

rafaelbn:
Please @gurneyalex as this PR will be necessary for more PR as:
- https://github.com/OCA/project-reporting/issues/9
- https://github.com/OCA/account-analytic/issues/36

could you review it?

In my opinion is very necessary to be near v9, as v9 already has this, making modules taking care of this fields will do that modules also useful in v9.

Maybe @StefanRijnhart and @dreispt are interested in this PR and could help us.

Thanks

pedrobaeza:
@rafaelbn Changes done

analytic_partner/models/account_analytic_line.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from openerp import models, fields
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    partner_id = fields.Many2one(
+        comodel_name='res.partner', string="Partner",
+        domain="[('parent_id', '=', False)]")
```
StefanRijnhart: If the rationale is to show only commercial partners, then you may want to filter on either top level partner ór 'is_company' == True.

StefanRijnhart:
Code looks good, but what is the idea behind this module? If it is to provide another axis, then why fill it with the partner of the analytic account (as does the follow up PR https://github.com/OCA/account-analytic/pull/36)?

rafaelbn:
All tested :+1: 
@StefanRijnhart idea is many partners in one analytic account and get info of analytic account by partner

StefanRijnhart:
@rafaelbn so why set the default to the analytic account's partner?

pedrobaeza:
@StefanRijnhart, as you have said, it's only a default value. This is for keeping consistency when an analytic account has the partner informed, and it's selected. It brings the associated partner, but you can then select another one if you want. Do you propose any other thing?

StefanRijnhart:
It seems to me that by setting it to the partner of the analytic account it can be easily mistaken for a related field. I think the name and label of the column should indicate that this is not the first axis partner of the analytic account, and I think it would be more clear if the default value was empty.

analytic_partner/models/account_analytic_line.py
rafaelbn: @pedrobaeza have you review this?

> 'is_company' == True

rafaelbn:
Hi @StefanRijnhart , I understand what you say... I would be very clear two columns one for analytic account's partner and second for analytic account line's partner... But we try to do same as Odoo v9.. Tested in runbot v9 http://93659-master-qsm-7129a9.runbot.odoo.com:
- Odoo add this field
- Odoo set (if established) in on change the partner from account to line.

Why Odoo make this? In my opinion because if you have an analytic account with a partner defined you will not want to change the partner in the account analytic line by default, you will change it manually.

Thats why we do like this.

Moreover (usability) if default value is empty, user could mistake an account analytic for other partner. Example: Project (one partner A), Project (no partner - multiple). If you select Project and it appears Partner A, user will know that he/she mistakes because wanted Project (for multiple) and Partner Z.

StefanRijnhart:
@rafaelbn Thanks for the pointer to Odoo 9.0. There, it is in fact a related field: https://github.com/odoo/odoo/blob/master/addons/analytic/models/analytic.py#L348

rafaelbn:
Yes @StefanRijnhart  because Odoo thinks that you must create a account analytic for each partner. And this is very limited.

pedrobaeza:
OK, let's see if this solution makes happy everyone: I have added the same related field to the partner of the analytic account and call it "Account Partner", and let the other specific partner too. Account Partner is readonly to avoid the false pretension that the data is only changed for that record.

StefanRijnhart:
Thanks for taking my arguments into account. However, because partner_id on the line is the related field to the account's partner in 9.0, I would prefer if you could rename your independent 'partner_id' field to something like 'other_partner_id'. You don't need to keep the related field introduced in revision 9f60dd, but if you do you could rename it to 'partner_id' as a proper backport of 9.0.

pedrobaeza:
@StefanRijnhart, I have taken into account your remarks, making partner_id the related one, renaming the other to other_partner_id, and changing labels accordingly. Please review it to merge this one as soon as possible as this one is the dependency for others that I have to publish.

gurneyalex:
:+1:

StefanRijnhart:
Thanks a dozen! :+1:

PR #40, opened 2015-09-28
[ADD] migrate account_analytic_project

hbrunn:


pedrobaeza:
I think that this situation has changed now in v8, because if you have created an analytic account, you only need to check "Tasks" field for having automatically created the project. Do you think this module is still worth in this case?

hbrunn:
@NL66278 what do you think?

NL66278:
@pedrobaeza @hbrunn I just tested creating a project by checking the Tasks fields, and this worked quite well. Only missing direct link between analytic account and the actual project.

NL66278:
@hbrunn @pedrobaeza So I think the module is still usefull, although I now would be able to live without it.

hbrunn:
and it's weird to have to turn on tasks for creating a project... I just looked into the code in order to reuse the create-the-project-function, but well, see for yourselves: https://github.com/OCA/OCB/blob/8.0/addons/project/project.py#L1204

sbidoul:
This module makes sense to me.
In particular fields `has_project` and `project_ids` are very useful by themselves.

NL66278:
:+1:

pedrobaeza:
OK then, go ahead.

account_analytic_project/README.rst
```diff
@@ -0,0 +1,53 @@
...
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/87/8.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/account-analytic/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/{project_repo}/issues/new?body=module:%20account_analytic_project%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
adrienpeiffer: Replace `{project_repo}`

account_analytic_project/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from . import model
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
adrienpeiffer: Remove this line

account_analytic_project/__openerp__.py
```diff
@@ -0,0 +1,35 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Project and analytic account integration',
+    'version': '1.0',
```
adrienpeiffer: Replace with the new version format `8.0.1.0.0`

account_analytic_project/__openerp__.py
```diff
@@ -0,0 +1,35 @@
...
+    'author': 'Therp BV',
```
adrienpeiffer: Add OCA as author

account_analytic_project/model/account_analytic_account.py
```diff
@@ -0,0 +1,78 @@
...
+    def create_project_for_account(self):
+        """Create a project for an already existing analytic account."""
+        project_model = self.env['project.project']
+        for this in self:
+            if not this.has_projects and this.type != 'view':
+                project_model.create({
+                    'name': this.name,
+                    'analytic_account_id': this.id,
+                })
+        return True
```
adrienpeiffer: is it necessary ?

adrienpeiffer:
I just have a little question. the "View Projects" button is present on the analytic account form as soon as "has_projects" is true and the "Create Project" button is hidden. Therefore, it is not possible to create multiple projects at the same analytical account ? Technically, I fully understand the use of a One2Many field to use the inverse relationship of many2one field present on the project. But maybe we should rename the button label for "View Project" or allow to create several projects for the same analytic account if it makes sense.

account_analytic_project/model/account_analytic_account.py
NL66278: Yes

NL66278:
@adrienpeiffer Thanks for your review. I will correct things like the README.rst. On the question of multiple projects for one analytic account: I think from a functional point of view, this is not desired and would create confusion. I only take into account the possibility that multiple projects might exist where absolutely necesarry.

adrienpeiffer:
Yes, technically it's possible, but for now, I do not see how a user can do this. So I don't really understand the label "View Project**S**" What do you think about that ?

account_analytic_project/model/account_analytic_account.py
adrienpeiffer: Why ?

hbrunn:
@adrienpeiffer thanks, I took your points and leave the rest of the discussion to @NL66278

account_analytic_project/model/account_analytic_account.py
NL66278: To provide the code for the button to create a project? Not really intuitive to check "use tasks" in order to indirectly create a project. And you might want a project, but no tasks.

hbrunn: I think @adrienpeiffer's question refers to the `return True`. In earlier version of OpenERP, you got an annoying error message when buttons didn't return anything, but by now, it doesn't matter to my knowledge (which is why I'm quite indifferent about keeping this line or removing it)

NL66278: @hbrunn @adrienpeiffer Actually I did not know that returning True was no longer required. If the question was about this line, I see no problem - on the contrary - in removing it.

adrienpeiffer: @hbrunn @NL66278 Yes it's was about the `return true`

account_analytic_project/__openerp__.py
```diff
@@ -0,0 +1,35 @@
...
+    'version': '8.0.1.0.0',
+    'author': 'Therp BV,Odoo Community Association (OCA)',
+    'category': 'Generic Modules/Accounting',
+    'website': 'https://therp.nl',
+    'email': 'info@therp.nl',
+    'depends': [
+        'account',
+        'project',
+    ],
+    'update_xml': [
```
adrienpeiffer: Use `data` instead of `update_xml` To avoid to have a red cross on runbot CI

hbrunn:
@adrienpeiffer thanks for your remarks, done & squashed commits

account_analytic_project/__init__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: UTF-8 -*-
+"""Import python code for module."""
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2015 Therp BV <http://therp.nl>
+#    All Rights Reserved
```
lmignon, line 7: you can remove 'All Rights Reserved' from all your licence headers.

account_analytic_project/model/__init__.py
```diff
@@ -0,0 +1,23 @@
+# -*- encoding: utf-8 -*-
+"""Import modules to extend models."""
```
lmignon, line 2: Do we need this comment?

account_analytic_project/model/account_analytic_account.py
```diff
@@ -0,0 +1,77 @@
+# -*- encoding: utf-8 -*-
+"""Extend account.analytic.account."""
```
lmignon, line 2: Do we need this comment?

lmignon:
@hbrunn Thank your for this module.
Few nitpicking comments otherwise :+1:

adrienpeiffer:
@hbrunn @NL66278 The Travis build was green before the last rebase but now it is red ...
Any idea?

NL66278:
It seems to me the error is totally unrelated to the changes proposed here. Something going wrong with multicurrency: 2015-10-01 08:08:39,642 22859 ERROR openerp_test openerp.addons.analytic_multicurrency.tests.test_analytic_lines: ` AssertionError: 0 != 200.0 within 7 places
And two more errors in the same module.

account_analytic_project/model/account_analytic_account.py
NL66278, line 2: Pylint requirement

lmignon, line 2: Requirement in your dev env? In all your code, this kind of comment appears only twice. I'll not block for this comment and you already have my :+1: . It's just a matter of consistency and IMO this kind of comment doesn't provide an added value.

account_analytic_project/__init__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: UTF-8 -*-
+"""Import python code for module."""
```
dreispt, line 2: Why this docstring?

account_analytic_project/model/account_analytic_account.py
dreispt, line 2: I agree with @lmignon: not a blocking issue, but these docstrings come unexpected and I would prefer to not have them

account_analytic_project/model/account_analytic_account.py
```diff
@@ -0,0 +1,77 @@
...
+                })
+
+    @api.multi
+    def view_projects_for_account(self):
+        """View list or form for project(s) related to analytic account."""
+        self.ensure_one()
+        project_ids = self.project_ids
+        assert project_ids, (_(
+            'This method is only valid for an analytic account having projects'
+        ))
```
dreispt: Not an issue, but translation is not really needed for assert messages

dreispt:
It is expected to have 1 Analytic <-> 1 Project, although the data model can support 1 Analytic <-> n Projects. So I prefer to keep "View Project" and not change "View Projects". The code is ready for the unexpected but possible case where 1 analytic has many projects, and that's good.

Some minor notes, but :+1:

account_analytic_project/model/account_analytic_account.py
sbidoul, line 2: @NL66278 You mean MQT now requires a module-level docstrings? Strange to me. But I did not follow recent MQT developments.

NL66278, line 2: MQT says we should follow pep standards, to be checked with pylint. Pep-0257 says "All modules should normally have docstrings", and pylint gives a missing-docstring error when it is not present. But then pylint will not complain when run in travis, because in the OCA setup only a small subset of checks has been enabled. I think here: travis/cfg/travis_run_pylint.cfg

lmignon:
It seems that this one is ready to be merged.

pedrobaeza:
Not really. There are some comments not attended and Travis is failing.

NL66278:
I made a new pull request #44  taking @hbrunn  's branch, rebasing it on current 8.0 and adressing some of the remarks,

sbidoul:
Superseded by #44

PR #46, opened 2015-11-30
[ADD] Add pos_analytic_config module

adrienpeiffer:
Use analytic account defined on POS configuration for invoices and accounting entries generated form POS orders.

cc\ @legalsylvain @elicoidal @pedrobaeza @Cedric-Pigeon


adrienpeiffer:
Supersedes https://github.com/OCA/pos/pull/65

legalsylvain:
:+1: Thanks.

pos_analytic_config/README.rst
```diff
@@ -0,0 +1,50 @@
...
+    :alt: License: AGPL-3
+
+===================
+POS Analytic Config
+===================
+
+This module allows to use analytic account defined on POS configuration
+for invoices and accounting entries generated from POS orders.
+
+Usage
```
elicoidal: I miss a Configuration section with steps to set up and screenshot.

pos_analytic_config/README.rst
```diff
@@ -0,0 +1,50 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+===================
+POS Analytic Config
```
elicoidal: POS Analytic Account

pos_analytic_config/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
```
elicoidal: header

elicoidal:
minor modifications. Otherwise LGTM (no test, no code review)

pedrobaeza:
I still think that the module should be renamed to _pos_analytic_.

pos_analytic_config/README.rst
```diff
@@ -0,0 +1,50 @@
...
+
+This module allows to use analytic account defined on POS configuration
+for invoices and accounting entries generated from POS orders.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/184/8.0
```
pedrobaeza: Change to 87 (this repo id)

pedrobaeza:
Apart from this, the rest LGTM.

elicoidal:
agree with pos_analytic

pos_analytic_config/models/__init__.py
adrienpeiffer: Done

adrienpeiffer:
Thanks @elicoidal and @pedrobaeza !
Concerning the module name, I do not really agree. I agree with the PoV of @legalsylvain on https://github.com/OCA/pos/pull/65 . This is only a way to implement the analytic in the pos in the configuration of points of sales.

pedrobaeza:
But my PoV is that you can join all these approaches in the same _pos_analytic_ module. And the end, you'll get a conflict of which one has the priority, so it's better to joint them in a module and solve this preference in it than splitting in a lot of mini-modules. We can put in "Known issues / Roadmap" that there can be other approaches for filling the analytic account.

legalsylvain:
Ok we are all agree that there are many ways to implements analytic on PoS. Good news.

The question is one module per approach (mini modules) or one module for many approaches.

As the @adrienpeiffer module approach is the more generic (one analytic account for all PoS sales), the module could be indeed renamed into 'pos_analytic' and receive after other PR to make it more specific. (that is theory).

In practice, other way to implement analytic for PoS will require other extra modules. (account_analytic_default for example as mentionned on https://github.com/OCA/pos/pull/65).
- Extra dependencies + keep 1 module = install unnecessary modules; 
- Other solution : create many module : 
  -> pos_analytic_config
  -> pos_analytic_by_product ...

My 2 cents.

elicoidal:
I am in favor of better granularity even if that means installing 2 modules.

lmignon:
I'm also in favor of better granularity. +1 to keep `pos_analytic_config`

sbidoul:
How about `pos_analytic_by_pos_config`?

pedrobaeza:
I like the last name, but without repeating the word pos: `pos_analytic_by_config`

adrienpeiffer:
Everyone agrees with `pos_analytic_by_config` ?

sbidoul:
:+1: for pos_analytic_by_config

adrienpeiffer:
@pedrobaeza @sbidoul @lmignon @elicoidal @legalsylvain @Cedric-Pigeon Now, it's `pos_analytic_by_config`. Thanks !

pedrobaeza:
:+1:

sbidoul:
I think this one can be merged.

legalsylvain:
indeed.

PR #52, opened 2016-02-04
[ADD][8.0] stock_analytic

andhit-r:
Move https://github.com/OCA/stock-logistics-workflow/pull/179 into OCA/account-analytic. Preserve commit history

PR #53, opened 2016-02-04
[ADD][8.0] stock_analytic

andhit-r:
Move stock_analytic into OCA/account-analytic. The module retrieve from https://github.com/julius-network-solutions/julius-openobject-addons and https://github.com/ClearCorp-dev/stock-logistics-workflow.

Commit history preserved from both branch. Several merge conflict had been resolved.


stock_analytic/README.rst
```diff
@@ -0,0 +1,33 @@
...
+Adds an analytic account in stock move to be able to get analytic information
+when generating the journal items.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Fabio Vílchez <fabio.vilchez@clearcorp.co.cr>
```
pedrobaeza, line 69: Add here Julius people and you

pedrobaeza:
Same comments as in https://github.com/OCA/account-analytic/pull/52

andhit-r:
@pedrobaeza i'm not confident regarding my retrieving technique. Is it safe to proceed?

pedrobaeza:
Yeah, I think so. It seems totally correct.

andhit-r:
@pedrobaeza I noticed that this module depend on **purchase** and **sale** module. Is it wise to depend on those modules since we only need **stock_account** to give analytic account capability to stock.move?

Maybe I could create **stock_sale_analytic**, **stock_purchase_analytic**, and maybe **stock_mrp_analytic** to achieve modularity.

pedrobaeza:
For what is that modules needed? If there's no other choice, to keep modularity it would be good to split it.

andhit-r:
I mean this original module will **force users** to install sale and purchase. I think the main functionality for this module **only to give analytic account on stock.move**. So when stock.move create account.move.line it will automatically fill analytic account field on account.move.line. So no need to depend on sale and purchase.

**stock_sale_analytic**
depend: stock_analytic, sale_stock
functionality: automatically give analytic account on stock.move according to project_id field on sale.order

**stock_purchase_analytic**
depend: purchase, stock_analytic
functionality: automatically give analytic account on stock.move according to account_analytic_id field on purchase.order

I don't like to force users install modules that they don't need. But if my suggestion is somehow overkill I will continue as it is.

pedrobaeza:
No, of course is a good proposition, but I was asking for the reason to need sale or purchase modules. Now I understand that is to propagate the corresponding analytic accounts, but I agree that that should be done in the modules you propose. Maybe a better name should be _sale_stock_analytic_ y _purchase_analytic_ to follow Odoo's naming convention.

andhit-r:
Ok @pedrobaeza. Thanks for your input. I will finish this module first.

andhit-r:
Travis didn't run again

pedrobaeza:
Seeing https://www.traviscistatus.com/, they experimented a problem and now they have some troubles to get up to date with all the checks, so wait a bit or try again to resubmit.

andhit-r:
close to restart travis

andhit-r:
@pedrobaeza would you kindly give need review label on this PR. Thanks before.

pedrobaeza:
Done

andhit-r:
Thanks @pedrobaeza.

andhit-r:
Hi @OCA/accounting-maintainers would you kindly review this PR. Many thanks

pedrobaeza:
:+1:

andhit-r:
Thanks @pedrobaeza

andhit-r:
Hi @OCA/accounting-maintainers would you kindly review this PR. Many thanks

JordiBForgeFlow:
@aheficent and me will review.

andhit-r:
Thank you @jbeficent

aheficent:
IMHO I think It should create the analytic entries for income or expenses account only. It seems to me that it's creating always the analytic entries for any accounting entry: https://github.com/andhit-r/account-analytic/blob/8.0-stock_analytic_1/stock_analytic/models/stock.py#L36 is creating. Do you agree with me @andhit-r ?
cc @jbeficent

stock_analytic/models/stock.py
```diff
@@ -0,0 +1,43 @@
...
+        res = super(StockQuant,
+                    self)._prepare_account_move_line(
+                        move, qty, cost,
+                        credit_account_id,
+                        debit_account_id,
+                        context=context
+                        )
+
+        # Add analytic account in debit line
+        if move.account_analytic_id:
```
aheficent: IMHO I think It should create the analytic entries for income or expenses account only. This creates two analytic entries, one for the debit and one for the credit, even for assets accounts. Do you agree with me @andhit-r ?
cc @jbeficent

stock_analytic/models/stock.py
```diff
@@ -0,0 +1,43 @@
...
+            res[0][2].update({
+                'analytic_account_id': move.account_analytic_id.id,
+            })
+            res[1][2].update({
+                'analytic_account_id': move.account_analytic_id.id,
+            })
+        return res
```
AaronHForgeFlow, line 45: IMHO I think It should create the analytic entries for income or expenses account only. This creates two analytic entries, one for the debit and one for the credit, even for assets accounts. Do you agree with me @andhit-r ?
cc @jbeficent

AaronHForgeFlow:
What else is needed to get this merged?

pedrobaeza:
Get more reviewers. You haven't approved the PR yet and there's a question unanswered as I see.

stock_analytic/models/stock.py
andhit-r: @aheficent agree. I push a commit to correct this. Analytic account will only be assigned if account != valuation account

andhit-r: Sorry for really-really late reply. Kindly review this again. Many thanks.

andhit-r:
Try to restart travis

andhit-r:
Travis fails because of other modules

andhit-r:
@jbeficent @aheficent all checks green. Would you kindly review it again. Many thanks.

@mikevhe18 @azmimr67 please help review

stock_analytic/models/stock.py
AaronHForgeFlow: Thank you @andhit-r I like the proposed changes. There's a problem with runbot, so i will on local. The code looks correct to me.

stock_analytic/models/stock.py
```diff
@@ -0,0 +1,45 @@
...
+# © 2016 Andhitia Rama <andhitia.r@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import fields, models, api
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    account_analytic_id = fields.Many2one(
```
AaronHForgeFlow: IMHO i think that the name of the field should be analytic_account_id. I think the name of the field should match the name of other modules as for example the account module https://github.com/OCA/OCB/blob/8.0/addons/account/account.py#L2398 or the project module https://github.com/OCA/OCB/blob/8.0/addons/project/project.py#L231

AaronHForgeFlow (commented):
I tested on local and reviewed the code and it works correctly. The only issue is related with the name of the field. What do you think about this?

AaronHForgeFlow (approved):
Tested on runbot + code review :+1: 

Thank you @andhit-r and @mikevhe18 Great job!

mikevhe18:
:+1: tested on runbot

PR #55, opened 2016-04-07
[ADD] module to manage analytic on purchase order

laetitia-gangloff:
This module allow user to set analytic account on purchase order instead of purchase order line.
When analytic account is set on purchase order, all lines of this purchase order will be on this analytic account.

### DIFF
--- a/purchase_analytic/models/purchase.py
+++ b/purchase_analytic/models/purchase.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+# © 2016  Laetitia Gangloff, Acsone SA/NV (http://www.acsone.eu)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    project_id2 = fields.Many2one(
+        comodel_name='account.analytic.account',
+        string='Contract / Analytic',
+        help='Use to store the value of project_id if there is no lines')
+    project_id = fields.Many2one(
+        compute='_compute_project_id',
+        inverse='_inverse_project_id',
+        comodel_name='account.analytic.account',
+        string='Contract / Analytic', readonly=True,
+        states={'draft': [('readonly', False)]},
+        store=True,
+        help="The analytic account related to a sales order.")
+
+    @api.multi
+    @api.depends('order_line.account_analytic_id')
+    def _compute_project_id(self):
+        """ If all order line have same analytic account set project_id
+        """
+        for po in self:
+            al = po.project_id2
+            if po.order_line:
+                al = (po.order_line[0].account_analytic_id) or False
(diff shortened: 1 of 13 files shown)


purchase_analytic/models/purchase.py
```diff
@@ -0,0 +1,66 @@
...
+    project_id = fields.Many2one(
+        compute='_compute_project_id',
+        inverse='_inverse_project_id',
+        comodel_name='account.analytic.account',
+        string='Contract / Analytic', readonly=True,
+        states={'draft': [('readonly', False)]},
+        store=True,
+        help="The analytic account related to a sales order.")
+
+    @api.one
```
andhit-r: @api.one deprecated

purchase_analytic/models/purchase.py
```diff
@@ -0,0 +1,66 @@
...
+        al = self.project_id2
+        if self.order_line:
+            al = (self.order_line[0].account_analytic_id) or False
+            for ol in self.order_line:
+                if ol.account_analytic_id != al:
+                    al = False
+                    break
+        self.project_id = al
+
+    @api.one
```
andhit-r: @api.one deprecated

andhit-r:
@laetitia-gangloff No unittest for onchange?

laetitia-gangloff:
@andhit-r  thank for your review

There is a test on onchange on purchase order line : https://github.com/acsone/account-analytic/blob/8.0-purchase_analytic-lga/purchase_analytic/tests/test_purchase_analytic.py#L24
And one on project_id change : https://github.com/acsone/account-analytic/blob/8.0-purchase_analytic-lga/purchase_analytic/tests/test_purchase_analytic.py#L50

Is it not sufficient ?

andhit-r:
@laetitia-gangloff sorry i guess i missed it. Other than @api.one usage :+1:

laetitia-gangloff:
Thank you !

I update @api.one in @api.multi

yvaucher:
@laetitia-gangloff What about purchase_analytic_global module which existed in 7.0 ?

https://github.com/OCA/purchase-workflow/tree/7.0/purchase_analytic_global

Which I ported to 9.0: https://github.com/OCA/purchase-workflow/pull/228 (just found out this PR as it was not in the same repository)

laetitia-gangloff:
I miss it :(

feketemihai:
@laetitia-gangloff I think the design is quite complicated, if you want to ease up, why not leave one single field for analytic, add an onchange on it (to change all lines with that value), and the field in context of purchase order lines, if you want to change it for some line, you can still do it...but having an inverse from purchase order lines to purchase order it doesn't make any sense..

AaronHForgeFlow:
Maybe an onchange method would be enough in this case. I'm not against of refactoring this module but this is working fine from a purely functional perspective.

AaronHForgeFlow:
Closing due to inactivity. Please reopen it if you plan to finish this. Regards.

PR #59, opened 2016-06-07
[9.0][PORT] account_analytic_required

yvaucher:


account_analytic_required/tests/test_account_analytic_required.py
```diff
@@ -1,85 +1,86 @@
...
 from openerp.tests import common
-from openerp.osv import orm
+from openerp import exceptions
 
 
-class test_account_analytic_required(common.TransactionCase):
+class TestAccountAnalyticRequired(common.TransactionCase):
 
     def setUp(self):
         super(test_account_analytic_required, self).setUp()
```
cyrilgdn: s/test_account_analytic_required/TestAccountAnalyticRequired/

So Travis is not happy.

account_analytic_required/models/account.py
```diff
@@ -0,0 +1,72 @@
...
+                              move_line.account_id.name,
+                              move_line.name,
+                              move_line.analytic_account_id.code,
+                              move_line.analytic_account_id.name)
+
+    @api.multi
+    def _check_analytic_required(self):
+        return not self._check_analytic_required_msg()
+
+    _constraints = [(_check_analytic_required,
```
guewen: is there a reason for not using `@api.constrains`?

guewen:
Fixed the class name here: https://github.com/yvaucher/account-analytic/pull/1

account_analytic_required/models/account.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @api.model
+    def _get_analytic_policy(self, account):
+        """ Extension point to obtain analytic policy for an account """
+        return account.user_type_id.analytic_policy
+
+    @api.multi
+    def _check_analytic_required_msg(self):
+        for move_line in self:
+            if move_line.debit == 0 and move_line.credit == 0:
```
alexis-via: Maybe we should use float_is_zero here...

alexis-via:
same remarks as guewen about the use of api.constrains ; otherwise :+1:

sbidoul:
@yvaucher do you plan to do the final touches to this one? (api.constrains + float_is_zero)

yvaucher:
@sbidoul will work on it this week

yvaucher:
fixed.

Travis is failing due to https://github.com/OCA/pylint-odoo/issues/48

sbidoul:
Is it just me or there is no menu to manage account types?

sbidoul:
Anyway, :+1: (code review and functional test on runbot).

Thanks for this work!

yvaucher:
@sbidoul indeed there is no menuitem to access the view for account types and the field account type on the account is with widget type selection thus you have no way to enter an account type without creating the menuitem. Would it be better to restore the menuitem in this module or is it a requirement we can see in other modules and thus we need it in a separate module ?

sbidoul:
@yvaucher we can leave that menu to another module if needed. I see a case for not letting end user fiddling with account types.

alexis-via:
Can someone with the super-powers merge this ?

rafaelbn:
:open_mouth: I would like to test it but I cannot find the menu por Account Types in v9 !

sbidoul:
@rafaelbn there is no menu indeed. See https://github.com/OCA/account-analytic/pull/59#issuecomment-232082642 above.

To test, you can manually change the account.account view to remove widget="selection" on user_type, so you get a way to access the account.type form view.

rafaelbn:
Sorry. Tested :+1:

account_analytic_required/README.rst
```diff
@@ -0,0 +1,63 @@
...
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/account-analytic/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Author
```
pedrobaeza: You should remove this distinction. Everyone is a contributor. Alexis was the first contributor of a long list.

account_analytic_required/README.rst
```diff
@@ -0,0 +1,63 @@
...
+------------
+* Stéphane Bidoul
+* Sefan Rijnhart
+* Laetitia Gangloff
+* Luc De Meyer, Noviat <info@noviat.com>
+* Yannick Vaucher <yannick.vaucher@camptocamp.com>
+
+Maintainer
+----------
+.. image:: http://odoo-community.org/logo.png
```
pedrobaeza: https (in all places)

account_analytic_required/models/account.py
```diff
@@ -0,0 +1,74 @@
...
+             "there is no analytic account ; if you select 'Never', the "
+             "accountant will get an error message if an analytic account "
+             "is present.")
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    @api.model
+    def _get_analytic_policy(self, account):
```
pedrobaeza, line 38: Use a related field instead this ugly trick

account_analytic_required/models/account.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @api.model
+    def _get_policies(self):
+        """This is the method to be inherited for adding policies"""
+        return [('optional', 'Optional'),
+                ('always', 'Always'),
+                ('never', 'Never')]
+
+    analytic_policy = fields.Selection(
+        lambda self, *args, **kwargs: self._get_policies(*args, **kwargs),
```
pedrobaeza: You should put directly the list of tuples here, as new API allows to extend this list. Even if you use a method, it should be declared this way (not with lambda):

```
analytic_policy = fields.Selection(_get_policies, ...)
```

pedrobaeza:
We need a way to set the policy that this module adds:
- Add a menu here.
- Depends on other module that does it.

But we cannot merge this module as is.

sbidoul:
@pedrobaeza to me the module is perfectly usable as is because in most projects we provide account types through a data module and users don't change them.

pedrobaeza:
If you provide that data module, you need to depend on this one, or get a warning when it's not installed, so I don't see that as the correct option. One must be able to change this policy according business rules, not localization rules.

sbidoul:
I'm not saying an account types menu is not useful, or that it must be set by localisation or business rule or...

I just say this PR is useful as is, because I have a project where I use it right know, where the analytic policies are set in a data module, and I don't need a menu to allow the user to manage the analytic policies.

So to me there is no reason to block this PR.

pedrobaeza:
But it's for me, because this module has no value for me if we don't have that menu. That's why I'm asking for it.

adrienpeiffer:
:+1: (Code review and functionnal test)

rafaelbn:
Hi @yvaucher just a question, is it so complicated to add a menu to make it more useful for final users? Thanks

yvaucher:
@rafaelbn I'll do it once I'm there in my todo list I just didn't add one as I have no idea if this was a specific requirement for this module or it should have been done in a base module. But I'll do it there and I will let other do the extract if needed then.

rafaelbn:
Thanks for considering @yvaucher

account_analytic_required/models/account.py
yvaucher, line 38: I think this is there for inheritance purpose that's what I understood from "Extension point" in the docstring.

account_analytic_required/README.rst
yvaucher: I kept original credits as I don't want to change them without @alexis-via approval.

account_analytic_required/models/account.py
pedrobaeza, line 38: But now you don't need it with the new `selection_add` option

yvaucher, line 38: This is not a selection function you mix up with your other comment on lambda, right ?

yvaucher:
PR has been fixed and is ready.

I answered the 2 open comments.

account_analytic_required/README.rst
alexis-via: @yvaucher you can change this, no problem !

sbidoul:
:+1: 

@pedrobaeza @rafaelbn ok for you now?

rafaelbn:
Great, many thanks! :+1:

pedrobaeza:
:+1: Thanks

PR #61, opened 2016-06-12
[NEW] procurement_purchase_analytic: New module

carlosdauden:


JordiBForgeFlow:
Hi @carlosdauden The module that you propose is interesting, but mixes three modules. Sales, purchase, procurement.

The need for the analytic account in the procurement order is common in many projects that may not be necessarily connected with sales or purchases.

I suggest that you create 3 modules:
- procurement_analytic. See the module in v9 here: https://github.com/Eficent/service_management/tree/9.0/procurement_analytic
  I can make a PR. 
- purchase_procurement_analytic (depends on procurement_analytic)
- sale_procurement_analytic (depends on procurement_analytic)

That way you will achieve modularization and others will be able to reuse your work for other purposes.

carlosdauden:
Ok @jbeficent. I go to create 3 modules.

PR #66, opened 2016-07-21
[FIX] procurement_analytic: Field not show without purchase module

carlosdauden:


rafaelbn:
Tested, is fixed thanks! :+1: 

cc @sergio-teruel @pedrobaeza  @Tecnativa

procurement_analytic/views/procurement_analytic.xml
```diff
@@ -10,7 +10,8 @@
     <field name="inherit_id" ref="procurement.procurement_form_view"/>
     <field name="arch" type="xml">
         <field name="group_id" position="after">
-            <field name="account_analytic_id" groups="purchase.group_analytic_accounting"/>
+            <field name="account_analytic_id"
+                   groups="sale.group_analytic_accounting,purchase.group_analytic_accounting"/>
```
pedrobaeza, line 14: But sale is not a dependency of this module, isn't it? You should make the change in the sale_... module.

carlosdauden, line 14: You are right...
I close this PR

PR #67, opened 2016-07-27
9.0 Port product_analytic to v9

alexis-via:


alexis-via:
OK, Travis is green. The only drawback of this PR is that I forgot to get the git history of that module ; I copied the last version of the module of v8 and then ported it.

pedrobaeza:
Please respect the commit history following this guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0

alexis-via:
I will probably not have time for this myself before my holidays ; so if someone wants to propose another PR with git history, he can do it and it will replace this PR.

PR #69, opened 2016-09-05
[9.0][ADD] account_analytic_no_lines

ThomasBinsfeld:
# Account Analytic No Lines

This module hides analytics lines from accounting menus and disable their generation from an invoice or a move line.


sbidoul:
In many situations (ie when there are no analytic distributions, and when people dont consider timesheets as analytic entries), the generation of analytic entries for each account move is redundant, creates overhead and pose a risk of desynchronization between analytic data and GL entries. In such situations users are happy to do their analytic reporting on account move lines and don't need analytic entries.

This module disables the generation of analytic entries and related menus.

pedrobaeza:
Thanks for the explanation. Please include this rationale in the module description.

account_analytic_no_lines/README.rst
```diff
@@ -0,0 +1,60 @@
...
+the generation of analytic entries for each account move is redundant, creates overhead and poses a risk of desynchronization between analytic data and GL entries.
+In such situations users are happy to do their analytics reporting on account move lines and don't need analytic entries.
+This module disables the generation of analytic entries and related menus.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/{repo_id}/{branch}
```
pedrobaeza: Replace placeholders by their values

account_analytic_no_lines/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+    'depends': [
+        'account',
+        'analytic',
+    ],
+    'data': [
+        'views/account_analytic_account.xml',
+        'data/data_account_analytic_group.xml',
+        'views/account_analytic.xml',
+    ],
+    'demo': [
```
pedrobaeza: Don't put empty keys

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
```diff
@@ -0,0 +1,117 @@
...
+from openerp.tests.common import SavepointCase
+
+
+class TestAccountAnalyticNoLines(SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestAccountAnalyticNoLines, cls).setUpClass()
+
+        # ENVIRONEMENTS
```
pedrobaeza: Environments

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
```diff
@@ -0,0 +1,117 @@
...
+# Copyright 2016 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp.tests.common import SavepointCase
+
+
+class TestAccountAnalyticNoLines(SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
```
pedrobaeza, line 11: Use `self` instead of `cls` as normal Python convention.

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
```diff
@@ -0,0 +1,117 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp.tests.common import SavepointCase
+
+
+class TestAccountAnalyticNoLines(SavepointCase):
```
pedrobaeza, line 8: Why `SavepointCase` instead of `TransactionCase`?

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
```diff
@@ -0,0 +1,117 @@
...
+            'partner_id': cls.user.id,
+            'account_id': cls.account_440000.id,
+            'type': 'in_invoice',
+            'invoice_line_ids': [(6, 0, [cls.invoice_line_1.id,
+                                         cls.invoice_line_2.id,
+                                         cls.invoice_line_3.id])]})
+        cls.invoice.action_move_create()
+
+    def test_create_analytic_lines(self):
+        '''
```
pedrobaeza: Use double quota for documentation as stated in the PEP257

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
```diff
@@ -0,0 +1,117 @@
...
+        Test the expected result when the method ´create_analytic_lines` is
+        called on an invoice.
+        The expected result is that no analytic line has been created
+        for this invoice.
+        '''
+        move_lines = self.account_move_line.search([('invoice_id',
```
pedrobaeza: For using less lines:

```
move_lines = self.account_move_line.search([
    ('invoice_id', '=', self.invoice.id)])
```

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
```diff
@@ -0,0 +1,117 @@
...
+            line.create_analytic_lines()
+
+        analytic_lines =\
+            self.account_analytic_line.search([('move_id',
+                                                'in',
+                                                move_lines.ids)]).ids
+        self.assertEqual(analytic_lines, [])
+
+    def test_finalize_invoice_move_lines(self):
+        '''
```
pedrobaeza: Double quota

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
```diff
@@ -0,0 +1,117 @@
...
+                                                     '=',
+                                                     self.invoice.id)])
+        for line in move_lines:
+            line.create_analytic_lines()
+
+        analytic_lines =\
+            self.account_analytic_line.search([('move_id',
+                                                'in',
+                                                move_lines.ids)]).ids
+        self.assertEqual(analytic_lines, [])
```
pedrobaeza: Use `self.assertFalse`

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
```diff
@@ -0,0 +1,117 @@
...
+        '''
+        move_lines = self.account_move_line.search([('invoice_id',
+                                                     '=',
+                                                     self.invoice.id)])
+        self.invoice.action_move_create()
+        analytic_lines =\
+            self.account_analytic_line.search([('move_id',
+                                                'in',
+                                                move_lines.ids)]).ids
+        self.assertEqual(analytic_lines, [])
```
pedrobaeza: Use `self.assertFalse`

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
ThomasBinsfeld: According [PEP8](https://www.python.org/dev/peps/pep-0008/), `cls` must be the first argument of a class method.

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
ThomasBinsfeld: With `SavepointCase`, the setup is done once for all test cases and each test case is done in a transaction. I found it interesting here.

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
pedrobaeza: But this is not declared as `@classmethod`. That's the problem. See for example https://github.com/odoo/odoo/blob/06754b3237a44d8ba2a18bf407a3e65447ca390c/addons/account/tests/account_test_users.py#L8

ThomasBinsfeld: @pedrobaeza In this case the overridden method is the classmethod `setUpClass` not the instance method `setUp` 

https://github.com/odoo/odoo/blob/06754b3237a44d8ba2a18bf407a3e65447ca390c/openerp/tests/common.py#L181

pedrobaeza: And why are you using that one which is not the regular way? Please use standard methods.

lmignon: @pedrobaeza because it is standard from version 9.0. see https://github.com/odoo/odoo/blob/9.0/openerp/tests/common.py#L197 .

pedrobaeza: That are 2 different things: one is the class used, and another the setUp method used. @lmignon, in your example, the used method is the regular one. And about SavepointCase, in useless for this one, as there's only one test method, so no gain for now.

lmignon: Therefore what's the problem? At the moment you are right we don't take advantage of the use of `SavepointCase` because there's only one test. Nevertheless it is not a bad practice and it can not block the PR. Moreover we should add this method in the OCA guidelines as good practice.

pedrobaeza: Yeah, you're right. My confusion comes from using SavepointCase instead of TransactionCase. We can continue then with this (although in this case there's no performance improvement).
Sorry for the noise.

pedrobaeza:
:+1:

account_analytic_no_lines/tests/test_account_analytic_no_lines.py
lmignon: No stress :smirk: Thank you to you for your investment in the reviewing process of so much PR!!

ThomasBinsfeld: @pedrobaeza @lmignon Actually, there are two tests, that's why I choosed to use `SavepointCase`.

pedrobaeza: Right. Another reason more! Forget my transient mental lapsus

ThomasBinsfeld: @pedrobaeza  Thanks for the review

sbidoul:
@ThomasBinsfeld one more thing:

image

sbidoul (approved):
I just did a functional test of then last change.

adrienpeiffer (approved):
:+1:

adrienpeiffer:
This one works fine. I think it's ready to merge.

lmignon (approved):

PR #72, opened 2016-10-11
Migration of account_analytic_no_lines to 10.0

adrienpeiffer:


alexis-via:
I'm discovering this module just now... seems great !
I have a module on v8 called account_hide_analytic_line available on https://github.com/akretion/odoo-usability/blob/8.0/account_hide_analytic_line
Differences:
- your module goes further by not generating analytic lines
- my module goes further on BI by modifying the analytic analysis using account.entries.report instead of analytic.entries.report but giving the same result, cf the 2 last pieces of XML of this file:
  https://github.com/akretion/odoo-usability/blob/8.0/account_hide_analytic_line/account_view.xml

We could certainly improve your module with my module.

Otherwise :+1:

account_analytic_no_lines/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Account Analytic No Lines',
+    'summary': """
+        This module hides analytics lines from accounting menus and disable
```
pedrobaeza: Can we short this summary. It's a bit long and doesn't fit good in the modules view

pedrobaeza (approved):
Little comment, but everything seems good. Thanks for the tests.

adrienpeiffer:
@alexis-via as this one is merged in 9.0 I propose the keep this one as it and adds some improvement in another PR.

account_analytic_no_lines/__manifest__.py
adrienpeiffer: It's done :smile:

pedrobaeza:
@alexis-via, please use next time the squash and merge option to avoid a useless commit like "[CHG] Shorter summary". Please, it's very important to keep commit history clean for several reasons:
- Shorten the GIT download size
- Ease the log review
- When migrating modules, avoid a lot of commits from old branches (not yet this case, but soon...)

alexis-via:
Squashing commits is like cancelling accounting entries, it should be forbidden.
Download size is not an argument (OCA projects are very small compared to other projects) and I don't mind when there are a lot of commits... it just reflect the development work. For me, squashing a commit is a manipulation of the reality of the development of the code.

pedrobaeza:
@alexis-via. That's not true. I'm not saying squashing losing author identity, I'm saying squash dumb commits resulted from a review like this one: https://github.com/OCA/account-analytic/pull/72/commits/37a94f4bbfefc0d36aa739fc6d1d2cacb68ceadc. It doesn't reflect any development history that should be considered relevant. If you are not going to do it, then please don't touch the merge button and let me do it.

sbidoul:
Personally I don't mind having many commits as long as commit messages are reasonably explanatory. It does no harm and provide insights on the development history.

pedrobaeza:
You can check several references about the topic:
- http://jamescooke.info/git-to-squash-or-not-to-squash.html
- http://programmers.stackexchange.com/questions/263164/why-squash-git-commits-for-pull-requests
- http://blog.marc-andre.ca/2014/02/05/why-i-wont-squash-my-commits/. This one strengthen the idea of squashing when the commit history is not properly handled (or at in this case, if it comes after a review process).

At the end, it's also an opportunity to better document commits, that it's what I do when squashing and merging. Note also that if GitHub has added this option, it's for a reason. And finally, remember the pain when you migrate a module after the branch creation and the commit history is very long using the extraction method.

PR #87, opened 2017-04-19
[ADD] Analytic Dimensions Module

angelmoya:
# Analytic Tag Dimension

This module allows to group Analytic Tags on Dimensions.

Dimensions are created as custom field, then you can group by Dimensions on:
* Account/Adviser/Journal Items
* Account/Reports/Business Intelligence/Invoices
* Account/Reports/Business Intelligence/Analytic Entries

When you set Tags on Analytic Entries each custom fields for dimensions are updated with its Tag.

One Tag is only allowed on one Dimension, and you can not set more than one Tag from same Dimensions on Analytic Entry.

## Configuration

To configure this module, you need to:

- Go to /Accounting/Configuration/Analytic Accounting/Analytic Accounts Dimensions to create new Analytic Dimensions.

- Then you can create new Analytic Tags on this form, or go to /Accounting/Configuration/Analytic Accounting/Analytic Accounts Tags and set Dimension for each Tag.

## Known issues / Roadmap

* Analytic Entries with Tags created before install this module are not updated with its Dimensions.


pedrobaeza:
Very cool feature. I had the same implementation in mind, although we don't need it right now. I'll check it when having some spare time.

analytic_tag_dimension/README.rst
```diff
@@ -0,0 +1,70 @@
...
+
+======================
+Analytic Tag Dimension
+======================
+
+This module allows to group Analytic Tags on Dimensions.
+
+Dimensions are created as custom field, then you can group by Dimensions on Analytic Entries Report.
+
+When you set Tags on Analytic Entries each custom fields for dimensions are updated with its Tag. One Tag is only allow on one Dimension, and you can not set more than one Tag from same Dimensions on Analytic Entry.
```
elicoidal: One Tag is only allowed on one Dimension

analytic_tag_dimension/README.rst
```diff
@@ -0,0 +1,70 @@
...
+
+
+Configuration
+=============
+
+To configure this module, you need to go to /Accounting/Configuration/Analytic Accounting/Analytic Accounts Dimensions to create new Analytic Dimensions.
```
elicoidal: please use enumerated list for the procedure (with "#. "). It is easier to follow

analytic_tag_dimension/README.rst
```diff
@@ -0,0 +1,70 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+======================
+Analytic Tag Dimension
+======================
+
+This module allows to group Analytic Tags on Dimensions.
+
+Dimensions are created as custom field, then you can group by Dimensions on Analytic Entries Report.
```
elicoidal: In general try to have line no more than 80 characters to ease the review.

analytic_tag_dimension/models/analytic.py
```diff
@@ -0,0 +1,83 @@
...
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    @api.constrains('tag_ids')
+    def _check_field(self):
+        dimension_ids = self.tag_ids.mapped('analytic_dimension_id').ids
+        if len(self.tag_ids.ids) != len(dimension_ids):
+            raise ValidationError(
+                _("You can not set two tags from same dimension"))
```
elicoidal: add a period at the end of the sentence

analytic_tag_dimension/models/analytic.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 PESOL (http://pesol.es)
+#                Angel Moya (angel.moya@pesol.es)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, api, fields, _
```
elicoidal: s/openerp/odoo (everywhere)

analytic_tag_dimension/views/analytic_view.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
elicoidal: odoo

elicoidal (changes requested):
some nitpicking

angelmoya:
@elicoidal thanks for review, work is done

elicoidal (approved):

fclementic2c:
Hi @angelmoya 
We were on the verge to do a similar dev too :)

I have few comments: 
* A small bug: From the dimension tag form view, when you click on 'add an item' you cannot select an existing tag, you can only create a new one.
* I think we need a constraint to force to fill in the analytic account on invoices lines, move lines whenever you have tags belonging to a dimension on the same line...otherwise you can have tags but no analytical line created. What do you think?
* Add a contraint to prevent deletion of a analytic tag if already used  

What would be cool as improvements:
- Add the the field color index on the dimension form and get the tags inherited the color index so you have 1 color per dimension. Visually it is very convenient.
- Add a column dimension in the analytic tag form view
- Why not having this dimension on the account move line too ? The user could even ignore analytic lines (which are a nonsense in my view)

Thanks for your great job,

angelmoya:
Hi @fclementic2c 

> I have few comments:
>A small bug: From the dimension tag form view, when you click on 'add an item' you cannot select an existing tag, you can only create a new one.

It's not a bug, as tag_ids is one2many, the form widget works this way. 

If you want to set created tags you need to do from tag from.

>I think we need a constraint to force to fill in the analytic account on invoices lines, move lines whenever you have tags belonging to a dimension on the same line...otherwise you can have tags but no analytical line created. What do you think?

I think that many times the person who create invoice, and fill analytic tags (as in sale order) don't know the analytic account. Anyway this functionality could be implemented in other module.

>Add a contraint to prevent deletion of a analytic tag if already used

this functionality could be implemented in other module.

>What would be cool as improvements:
>Add the the field color index on the dimension form and get the tags inherited the color index so you have 1 color per dimension. Visually it is very convenient.

Nice feature, I will do it

>Add a column dimension in the analytic tag form view

I forgot to do, I wrote it on README... I will do it

>Why not having this dimension on the account move line too ? The user could even ignore analytic lines (which are a nonsense in my view)

I will check it, could be nice

>Thanks for your great job,

Thanks for review

fclementic2c:
:+1:

sbidoul:
Very nice contrib. 

Could we have the dimensions on the move lines too? 

Analytic lines where somehow useful with the old analytic distribution module, but nowadays, I see them as redundant with move lines (that's why we made account_analytic_no_lines).

angelmoya:
@sbidoul I will do on account_move_lines to.

But I think that account_move_lines is still usefull, at less for timesheet on projects.

angelmoya:
@sbidoul @fclementic2c  Analitic Dimensions added on account move lines

elicoidal:
Can you please check your travis?

analytic_tag_dimension/models/analytic.py
```diff
@@ -0,0 +1,118 @@
...
+        if len(self.tag_ids.ids) != len(dimension_ids):
+            raise ValidationError(
+                _("You can not set two tags from same dimension."))
+
+    @api.model
+    def create(self, values):
+        tag_values = values.get('tag_ids')
+        # tag_ids can not be created from here
+        if tag_values:
+            tag_ids = tag_values[0][2]
```
guewen: Are we guaranteed that `tag_values[0][2]` contains the ids? What if someones passes a `[(0, 0, {...})]` to the create/write? Maybe we just want to raise an error saying that only `[(6, 0, [ids])]` is supported in case?

angelmoya: I'm rewritting all, I was thinking about it, and now I'm doing with stored computed fields, instead of overwrite create and write function, and with abstract model to do it easiest.

angelmoya:
I'm rewritting all, I was thinking about it, and now I'm doing with stored computed fields, instead of overwrite create and write function, and with abstract model to do it easiest.

angelmoya:
I have not been able to do with computed stored field, instead of rewrite create and write function.

But I did this on abstract model, this way is easiest to implement on account.move.line and account.invoice.line.

For sale_order_line and purchase_order_line, I think that it's better to create new modules.

Also there are a bug on travis, I don't know why, if I use "with self.AssertRaises..." commit was done on database, so I can not create accounts on tests setup.

angelmoya:
I also update README.rst and implement analytic dimensions on account invoice report.

pedrobaeza:
I miss here to have a column for each dimension in the views on the affected models. This can be done overwriting fields_view_get method.

angelmoya:
Thanks @pedrobaeza I will try to do this week

angelmoya:
@pedrobaeza could you check this?, I create fields_view_get, but get an error on js

`Uncaught TypeError: Cannot set property '__attrs' of undefined`

Could be problem related with add fields on abstract model.

And could you send me an example to test this fields view get.

dreispt:
Sorry if this is a dumb question:
aren't Analytic Tags already supposed to be used for Dimensions ("Contract", "Geography", "Product line", ...) ?

dreispt:
I guess the answer to myself is that Tags are supposed to be "Parent Accounts" (although they don't provide the same features). So having a separate Dimension field is useful, since you can decide how to use Tags.

angelmoya:
The scope for this is to get better BI, for example you can get analytic like:

- Group by Partner/ Type of project / Project / Type of task

And you can use on diferents order like Group by Type of task by Partner

dreispt:
> The scope for this is to get better BI

A Analytic Account hierarchy, using Parent Account, already achieves that, and we don't strictly need dimensions for that.

AFAIK the "Dimension" concept is not to categorize AAs. It is rather to say that you should categorize the same line in several perpendicular complementary Dimensions. To be useful it needs to go along with a multiple AA distribution of the line.

For example:  for Dim1=Geography and Dim2=Product Line, we could have
- Sale 1: Dim1=Europe;Dim2=Desktops
- Sale 2: Dim1=Amercas;Dim2=Servers

pedrobaeza:
FYI we have also restored the multi-plan approach a bit modified on https://github.com/OCA/account-analytic/pull/90.

What I understood of this module is that let you write directly tags on the analytic lines, and thus, acting as a multi-plan replacement, but using tags instead of analytic accounts. Isn't it?

angelmoya:
No, it's not as multi-plan replacement.

On standard you have analytic tag on analytic lines, but as analytic tags is m2m field you can not use on BI.

With this module you can create dimensions, each dimension create automatically new field on analytic lines to suport group by this field.

You can see you analytic accounting as a plan if you order dimensions in one way, but you can order your group by to see your analytic accounting in several ways.... 

It's hard to explain, and my low english skills don't help me.

I will fix test to get available on runbot and prepare demo data to demo.

dreispt:
> each dimension create automatically new field on analytic lines to suport group by this field.

This sounds very good.

>  analytic accounting as a plan if you order dimensions in one way, but you can order your group by to see your analytic accounting in several ways

Yes, that exactly what multi-dimension are supposed to provide.

From the user PoV:

Managers want to perform multi dimension analysis, as you described for BI.
So they need each line to have available the value for each Dimension (Project, Resource Type, Market Segment, Geography, etc.).
Inside each Dimension we have a list of possible values, organized in a hierarchy or not.

Data Operators need assign the correct value for each Dimension when introducing some data: a timesheet line, a sale line, a PO line, etc.

I think what is confusing me here is that Tags are going to be used as alternate Analytic Accounts.
Why not using actual Analytic Accounts as an m2m on Lines, instead of Analytic Tags?
This would change very little in your implementation, and IMO it would provide a more solid model for analysis.

pedrobaeza:
@angelmoya what you are describing is indeed a multi-plan replacement...

angelmoya:
@pedrobaeza OK I was wrong about multi plan scope.

@dreispt I use tags instead of accounts because:
- account = project, and one task must be in one project
- we just have m2m to analytic_tag on analytic lines, invoice lines, account move lines...
- with abstract model is easy to create dimensions fields on each model that has analytic_tags as m2m

dreispt:
This isn't my preferred solution for the problem, but it's a reasonable solution.
I won't oppose it.

fclementic2c:
@dreispt Be careful not mixing up 'multi-dimension' and the breakdown of the cost WITHIN a dimension.

Do not mix also with the 'analytic distribution' which is a third thing... because distribution is done afterwards: at the end of the month or the year , for instance, to reallocate indirect cost (ex : ABC methods) whereas the breakdown (sorry I can't remember the correct terminology) is done at the time you post the journal entry.
Today we do not have analytic distribution tool in odoo.

So:
* multi-dimension is always 100% allocation on each dimension 
(see what I mean in this example: [here](https://docs.google.com/spreadsheets/d/1kUrH7_SGWsBXsDRNPEXP0UNRvYdIpkWyKACesS4xiUM/edit?usp=sharing) )
My point is that you cannot split for ex an amount 30% in BU1 and 70% on DepartmentB otherwise your pivot table (or group by in odoo) on these data will not do the math compares to general ledger.

* However, you can split the amount 30% in BU1 and 70% in BU2 because it belongs to the same dimension, this is the breakdown I am talking about.
This feature is very cool of course but it is more advanced and it could be an improvement of the multi-dimension that @angelmoya is proposing here

@pedrobaeza This is why I am not a big fan of the analytic distribution module : it does the job, true but nobody use it properly and nothing prevent you for doing wrong allocation. It is far too dangerous in my view and difficult to explain too... But of course you could add constraints to improve this.

pedrobaeza:
@fclementic2c thanks for the clarification. Of course, multi-plan, or the breakdown of the cost across multiple dimensions requires some constraints or teaching, and the current proposed module doesn't make the first, but for customers coming from 8.0 that are used to handle this, it's enough and they already have their hierarchical analytic chart with this multi-dimension approach in mind, That's why we migrate it and slightly improve it.

We can converge in any moment with another possible solution. I prefer the UI possibility to unfold the possible dimensions in the line instead of having a one2many/many2many embedded.

fclementic2c:
@pedrobaeza sure. I get your point. I am not even sure both can converge. Maybe customers will have to choose.

@angelmoya Can we merge this PR on the technical point of view ?

angelmoya:
@fclementic2c I need to fix or comment:
- account.invoice.report because sql view is not created propertly on database with data
- default view get, to create filter on search view

angelmoya:
@fclementic2c  done

PR #88, opened 2017-04-20
Add 10.0 project task analytic tag

ljimenezsidoo:
Project Task Analytic Tag
=========================

This module extends the functionality of project to support to
add Tags in the timesheet and allow you to use the tags of the projects in your timesheet

### DIFF
--- a/project_task_analytic_tag/README.rst
+++ b/project_task_analytic_tag/README.rst
@@ -1,0 +1,91 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=========================
+Project Task Analytic Tag
+=========================
+
+This module extends the functionality of project to support Tags in timesheet
+and the possibility to add them from the projects.
+
+The target to add analytic tags to project task is to create analytic lines
+(timesheet lines) with right tag.
+
+Also this module combine very well with the dimensions module
+(Analytic Tag Dimensions) for group analytic tags on dimensions.
+
+============
+Installation
+============
+
+To install this module, you need to:
+
+- First Add in the addons path the module and install in your Odoo.
+  And install the dependencies of the module
+
+=============
+Configuration
+=============
+
+To configure this module, you need to:
+
+- You only need to have all the dependencies installed.
+
+=====
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to project and enter in a task section,
+#. Go to the page timesheet
+#. You can add a new element.
+#. When you add the tag in the extra Information, the tag is automatically
+   set for newly added element.
+
+===========
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/project_task_analytic_tag/issues>`_. In case of trouble,
+please check there if your issue has already been reported. If you spotted it
+first, help us smash it by providing detailed and welcomed feedback.
+
+=======
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Luis Adan Jiménez Hernádez <luis.jimenez@pesol.es>
+* Ángel Moya <angel.moya@pesol.es>
+
+Funders
+-------
+
+The development of this module has been financially supported by:
+
+* PESOL
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/project_task_analytic_tag/__manifest__.py
+++ b/project_task_analytic_tag/__manifest__.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 PESOL (http://pesol.es)
+#                Luis Adan Jimenez Hernandez (luis.jimenez@pesol.es)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Project Task Analytic Tag",
+    "summary": "Analytic tag in project task",
+    "version": "10.0.1.0.0",
+    "author": "PESOL, Odoo Community Association (OCA)",
+    "website": "http://pesol.es",
+    "category": "Analytic Accont",
+    "license": "AGPL-3",
+    "depends": [
+        'project',
+        "hr_timesheet"
+    ],
+    "data": [
+        'views/project_task_analytic_view.xml',
+    ],
+    'installable': True,
+}
--- a/project_task_analytic_tag/models/project_task_tag.py
+++ b/project_task_analytic_tag/models/project_task_tag.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 PESOL (http://pesol.es)
+#                Luis Adan Jimenez Hernandez (luis.jimenez@pesol.es)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class ProjectTaskTag(models.Model):
+    _inherit = 'project.task'
+
+    analytic_tag_ids = fields.Many2many(
+        comodel_name='account.analytic.tag',
+        relation='project_task_analytic_tag_rel',
+        column1='analytic_tag_ids',
+        column2='default_tag_ids',
+        string='Analytic Tag')
--- a/project_task_analytic_tag/__init__.py
+++ b/project_task_analytic_tag/__init__.py
@@ -1,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 PESOL (http://pesol.es)
+#                Luis Adan Jimenez Hernandez (luis.jimenez@pesol.es)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from . import models
--- a/project_task_analytic_tag/models/__init__.py
+++ b/project_task_analytic_tag/models/__init__.py
@@ -1,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 PESOL (http://pesol.es)
+#                Luis Adan Jimenez Hernandez (luis.jimenez@pesol.es)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from . import project_task_tag
--- a/project_task_analytic_tag/tests/__init__.py
+++ b/project_task_analytic_tag/tests/__init__.py
@@ -1,0 +1,3 @@
+# -*- coding: utf-8 -*-
+
+from . import test_project_task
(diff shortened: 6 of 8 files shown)


project_task_analytic_tag/README.rst
```diff
@@ -0,0 +1,86 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=========================
+Project Task Analytic Tag
+=========================
+
+This module extends the functionality of project and mrp_repair to support to
```
elicoidal: mrp_repair?
It is not in dependency list and anyway would simply need another module if you'd wish to do so.

elicoidal: Isnot this module overlapping with https://github.com/OCA/project/tree/10.0/project_categ and https://github.com/OCA/project/tree/10.0/project_categ_issue?

I am not so much in favor of adding a new category system without clear value

angelmoya: Hi @elicoidal , you are right about README, it's a mistake about copy & paste from other module.

But this module is not overlapping with project category. The target to add analytic tags to project task is to create analytic lines (timesheet lines) with right tag.

This way you can set tags like "Design", "Develop", "Testing"... on your timesheet lines, and you will be able to get analytic about type of work you was doing. Also interesting in combination with analytic dimensions.

angelmoya:
I copy this comment because it's hidden by outdated change, but I think is usefull to see for other reviewers

>This module is not overlapping with project category. The target to add analytic tags to project task is to create analytic lines (timesheet lines) with right tag.
>This way you can set tags like "Design", "Develop", "Testing"... on your timesheet lines, and you will be able to get analytic about type of work you was doing. Also interesting in combination with analytic dimensions.

elicoidal:
@angelmoya then I would suggest to make it permanent in the README ;)

project_task_analytic_tag/README.rst
```diff
@@ -0,0 +1,85 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+Project Task Analytic Tag
```
elicoidal, line 6: Line before for the document title

project_task_analytic_tag/README.rst
```diff
@@ -0,0 +1,85 @@
...
+=========================
+
+This module extends the functionality of project to support to
```
elicoidal: "This module extends the functionality of project to support Tags in timesheet and the possibility to add them from the projects."

project_task_analytic_tag/README.rst
```diff
@@ -0,0 +1,85 @@
...
+add Tags in the timesheet and allow you to use the tags of the projects 
+in your timesheet
+
+Installation
+============
+
+To install this module, you need to:
+
+- First Add in the addons path the module and install in your odoo.
```
elicoidal: Odoo

project_task_analytic_tag/README.rst
```diff
@@ -0,0 +1,85 @@
...
+To configure this module, you need to:
+
+- You only need to have all the dependencies installed.
+
+Usage
+=====
+
+To use this module, you need to:
+
+- Go to project and enter in a task section, here you only need to go to the
```
elicoidal: make better use of enumerated list "#. " for procedures. For example:
```
#. Go to project and enter in a task section, 
#. Go to the page timesheet
#. You can add a new element. 
#. When you add the tag in the extra Information, the tag is automatically set for newly added element.
```

project_task_analytic_tag/README.rst
```diff
@@ -0,0 +1,85 @@
...
+  page timesheet and add a new element. if you put the tag in the
+  extra-Information, when you add a new element the tag is set auto.
+
+
+Known issues / Roadmap
```
elicoidal: remove the section

project_task_analytic_tag/README.rst
elicoidal: When i read the code, I dont get all above mentioned feature though.
Anything I missed?

elicoidal (changes requested):

project_task_analytic_tag/README.rst
elicoidal: Add a note in the README

project_task_analytic_tag/README.rst
angelmoya: Really no more code is needed, now on tasks, timesheets page are analytic entries (previously are timesheet_sheet_lines that are related).
Then this module just show tag_ids from analytic_account_line on project_task. And set analytic_tag_ids on project task to define default analytic_tag_ids for all timesheets (account_analytic_line) for this task.

ljimenezsidoo:
@elicoidal thanks for the comments all the changes updated

elicoidal (approved):

angelmoya (approved):

dreispt:
I wonder if OCA/project is not better suited for this.

From the functional PoV, it's not clear to me the use case for this.
Why do we want several analytic accounts on the same task? How do you analyse Tasks then?

angelmoya:
Hi @dreispt , we also were thinking about puting it on OCA/project , but finally we thought that final scope is to get correct analytic accounting.

You are wrong about this "several analytic accounts on the same task", it's not for analytic accounts, is for analytic tags. This way you can set type of work you are doing in one task.

Without this module you can not set analytic tags for this account.analytic.lines

dreispt:
@angelmoya I see, but what is the point. Can you give a real life example of haw this is used?
 Don't take me wrong, I just want to better understand the use case.

angelmoya:
Sorry @dreispt , I said you are wrong because you said account instead tag, and I thought that you can undertand the scope using analytic tags.

Ok, for example.

- you have a project, that is an analytic account, remember that you lose hierarchical structure on analytic account.
- then you create a task on this project, your task could be "Design 3D for new component" and you can set analytic tag on task "Design 3D". Other task could be "Manufacture prototype" with other analytic tag, on same project.
- as you set analytic tag on project task (then in timesheet lines), then your analytic account line has correct analytic tag.
- Then you can get from analytic account things like "Total amounts by Design 3D by project" or "Total by project by analytic tag (Design 3D, Prototype)"

My real use case is more easy, I improve the use of mrp_repair, to be able to create projects and task. Then all task created from RMA has analytic tag "RMA" and then you can get analytic account for RMA.

All this could be better if you install also #87 this way you can group analytic tags on dimensions like:
- type of project (Productive, RMA, ...)
- type of task (Design, Manufacture, Manage, ...)
- departments ....

dreispt:
For the example, IMO it would be better to have an Analytic Account on Tasks. This has been done on OCA/project (although looking the module name - `service_desk` - is unfortunate).
My use case is a "Corrective Maintenance" project, where I open tasks, and then on each Task I select the Contract (Analytic Account) it corresponds to.

pedrobaeza:
Since v9, there's a field analytic_account_id in tasks and issues...

dreispt:
@pedrobaeza Thanks! I guess it shows I'm still on v8 😛 
So we already know the task Analytic Tags - just follow the Task analytic_account_id, right?

pedrobaeza:
Yeah, at task or issue level you can predefine an analytic account to be filled for timesheet entries, so theoretically you can switch this analytic account for changing where to put the entry. We have been working on restoring analytic hierarchy. Have you seen it?

dreispt:
I haven't read the RFC yet, but I'm aware it is being discussed.
Tags don't replace Parents: they don't support things like WBS, relevant for non trivial Projects.

pedrobaeza:
It's no more an RFC, but an implementation: https://github.com/OCA/account-analytic/tree/9.0/account_analytic_parent and an improvement here: https://github.com/OCA/account-analytic/pull/91

angelmoya:
@dreispt with analytic dimension you can support things like WBS, and I think that it could be better than use analytic parent.

You can think on this structure with analytic parent:

- Customer 1
  - Project 1
    - Materials 1
    - Hours 1

With previous structure, if you use parent account you need to create one account for each Materials of each Project. And you need to set task on Materials or on Hours.

Also on BI you can not view amounts group rightly.

If you use dimensions you can do things like:

- Customer (directly customer because res.partner is on analytic.line)
  - Project (analytic account without parent)
   - Materials (dimension concept = materials, created dimanicaly on analytic.line)
   - Hours (dimension concept = hours, created dimanicaly on analytic.line)

And you can use on BI, and group on diferents dimensions on diferent order.

Is hard to me to explain in english, but I really think that tags with dimensions are better way to do analytic. And for BI is better than use parent account.

On v8 I did things like add parent_id on BI, and parent_parent_id, and product_category, to get similar functionality, and know this could be done automatically.

dreispt:
That was a great explanation, and I think I understand your point.
From my PoV the scenario you describe is implemented with two Dimensions: Dim1="Project" and Dim2="Resource"".
Both of them are a hierarchy of analytic accounts:

 - Project Dimension
    - Customer 1
        - Project 1
  - Resource Dimension
    - Materials
    - Time

Now each analytic line should have be assigned to a "analytic combination", with values for each of the Dimensions. have a look at [this](https://github.com/OCA/server-tools/pull/275#issuecomment-149817172).

Maybe what I describe if far from being possible in practice and your approach is a good short term solution. We're just discussing here.

angelmoya:
@dreispt @pedrobaeza one more review to aprove?

project_task_analytic_tag/models/project_task_tag.py
```diff
@@ -0,0 +1,17 @@
...
+
+from odoo import models, fields
+
+
+class ProjectTaskTag(models.Model):
+    _inherit = 'project.task'
+
+    analytic_tag_ids = fields.Many2many(
+        comodel_name='account.analytic.tag',
+        relation='default_tag_account_tag',
```
dreispt: Convention is for this table to end with `_rel`

project_task_analytic_tag/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "name": "Project Task Analytic Tag",
+    "summary": "Analytic tag in project task",
+    "version": "10.0.1.0.0",
+    "author": "PESOL",
+    "website": "http://pesol.es",
+    "category": "Analytic Accont",
+    "license": "AGPL-3",
+    "depends": [
+        'project',
+        "hr_timesheet"
```
dreispt, line 15: This is related to ` analytic_tag_dimension` right?
Does it make sense to add it as a dependency?
If not, at least that module should be mentioned in the README, as a way to use the information added here.

angelmoya: Hi @dreispt , it's not related, and is not a dependency.

You can use it alone, to set analytic tag on analytic lines created on timesheet.

We only talk about the other module to explain why we decided to improve analytic tags instead of migrate analytic account parent.

ljimenezsidoo: @dreispt The dependency is not necessary. As @angelmoya said you can use it alone.

ljimenezsidoo: @dreispt Do you agree that there is no need for dependency? I'm going to add in the readme file for more info as you said.

dreispt: ok

rafaelbn:
Rebuilding for testing @angelmoya

rafaelbn:
@angelmoya @acm1pt-colorado could you please rebase? I cannot test it. Travis fails and runbot doesn't build :cry:

angelmoya:
Done, I will check new travis error

rafaelbn:
Hi @angelmoya , I've tested functionally and it's :+1: 

But I don¡t understand why in this PR are you changing files of other modules: https://github.com/OCA/account-analytic/pull/88/files#diff-de1af174f37e99cfcd5af533f2eeb68f

:question: :smile:

rafaelbn (changes requested):
Functionally :+1: 

But in this PR "Files changed 934" :astonished:

angelmoya:
OK @rafaelbn , maybe was a problem with rebase, I did it again and do force push to do in one commit and only update my module files.

PR #97, opened 2017-05-11
[10.0][MIG] port purchase_analytic to V10.0

Cedric-Pigeon:
Migrate 8.0 PR to V10.0: https://github.com/OCA/account-analytic/pull/55

### DIFF
--- a/purchase_analytic/models/purchase.py
+++ b/purchase_analytic/models/purchase.py
@@ -2,7 +2,7 @@
 # © 2016  Laetitia Gangloff, Acsone SA/NV (http://www.acsone.eu)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp import api, fields, models
+from odoo import api, fields, models
 
 
 class PurchaseOrder(models.Model):
@@ -62,7 +62,3 @@ class PurchaseOrder(models.Model):
                 return
         self.project_id2 = self.project_id
         self.order_line = r
-
-    def _get_merge_order_key(self):
-        res = super(PurchaseOrder, self)._get_merge_order_key()
-        return res + ('project_id', )
--- a/purchase_analytic/models/purchase.py
+++ b/purchase_analytic/models/purchase.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+# © 2016  Laetitia Gangloff, Acsone SA/NV (http://www.acsone.eu)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    project_id2 = fields.Many2one(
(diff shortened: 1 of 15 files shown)


Cedric-Pigeon:
@gurneyalex Could you please add the need review label ?

JonathanNEMRY (approved):

JonathanNEMRY:
tested and backported into v9
@gurneyalex  I think it is more a `need review` than an `in progress` label :sweat_smile:

purchase_analytic/models/purchase.py
```diff
@@ -0,0 +1,64 @@
...
+        string='Contract / Analytic',
+        help='Use to store the value of project_id if there is no lines')
+    project_id = fields.Many2one(
+        compute='_compute_project_id',
+        inverse='_inverse_project_id',
+        comodel_name='account.analytic.account',
+        string='Contract / Analytic', readonly=True,
+        states={'draft': [('readonly', False)]},
+        store=True,
+        help="The analytic account related to a sales order.")
```
AaronHForgeFlow, line 22: "The analytic account related to a **purchase** order."

AaronHForgeFlow (approved):
:+1:  Code review

aheficent:
@Cedric-Pigeon  Could you add some tests?

feketemihai:
@Cedric-Pigeon I think first we should have the module merged in 8, and after to migrate it, check also my comment on #55 .

purchase_analytic/models/purchase.py
```diff
@@ -0,0 +1,64 @@
...
+# © 2016  Laetitia Gangloff, Acsone SA/NV (http://www.acsone.eu)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    project_id2 = fields.Many2one(
```
MiquelRForgeFlow, line 11: maybe it would be better to call it `manual_project_id`?

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
@feketemihai I don't think we should have to wait the merging of v8. @Cedric-Pigeon, could you attend comments?

AaronHForgeFlow:
@Cedric-Pigeon  Is is possible to attend comments here and move forward?

purchase_analytic/models/purchase.py
AaronHForgeFlow: From my side it is fine to keep the field name like this.

AaronHForgeFlow:
@Cedric-Pigeon can you rebase to check if ci is still green?

Cedric-Pigeon:
@aheficent finally done...

AaronHForgeFlow:
@pedrobaeza Can the issue with the CLA here block the Pull Request?

pedrobaeza:
No, that one can be ignored as former member of Acsone. We should provide somehow a way to bypass it, but it's not yet implemented

PR #98, opened 2017-05-12
FIX issue 96

angelmoya:
IMP analytic_dimension_id on account.analytic.tag not required

I do this field nos required, to prevent errors on app.

Anyway, IMO, create analytic tags without dimension could be wrong... but now you are free to work this way.

### DIFF
--- a/analytic_tag_dimension/models/analytic.py
+++ b/analytic_tag_dimension/models/analytic.py
@@ -44,21 +44,21 @@ class AccountAnalyticTag(models.Model):
 
     analytic_dimension_id = fields.Many2one(
         comodel_name='account.analytic.dimension',
-        string='Dimension',
-        required=True)
+        string='Dimension')
 
     @api.multi
     def get_dimension_values(self):
         values = {}
-        for tag in self:
+        for tag in self.filtered('analytic_dimension_id'):
             code = tag.analytic_dimension_id.code
             values.update({
                 'x_dimension_%s' % code: tag.id})
         return values
 
     def _check_analytic_dimension(self):
-        dimension_ids = self.mapped('analytic_dimension_id').ids
-        if len(self.ids) != len(dimension_ids):
+        tags_with_dimension = self.filtered('analytic_dimension_id')
+        dimensions = tags_with_dimension.mapped('analytic_dimension_id')
+        if len(tags_with_dimension) != len(dimensions):
             raise ValidationError(
                 _("You can not set two tags from same dimension."))
 


analytic_tag_dimension/models/analytic.py
```diff
@@ -44,21 +44,25 @@ class AccountAnalyticTag(models.Model):
...
         comodel_name='account.analytic.dimension',
-        string='Dimension',
-        required=True)
+        string='Dimension')
 
     @api.multi
     def get_dimension_values(self):
         values = {}
-        for tag in self:
+        for tag in self.filtered(
```
pedrobaeza: No need to put lambda. Just put `'analytic_dimension_id'`.

analytic_tag_dimension/models/analytic.py
```diff
@@ -44,21 +44,25 @@ class AccountAnalyticTag(models.Model):
...
                 'x_dimension_%s' % code: tag.id})
         return values
 
     def _check_analytic_dimension(self):
-        dimension_ids = self.mapped('analytic_dimension_id').ids
-        if len(self.ids) != len(dimension_ids):
+        tags_with_dimension = self.filtered(
+            lambda d: d.analytic_dimension_id
+        )
+        dimension_ids = tags_with_dimension.mapped('analytic_dimension_id').ids
```
pedrobaeza: No need to use .ids. Use directly recordsets.

analytic_tag_dimension/models/analytic.py
```diff
@@ -44,21 +44,25 @@ class AccountAnalyticTag(models.Model):
...
+        ):
             code = tag.analytic_dimension_id.code
             values.update({
                 'x_dimension_%s' % code: tag.id})
         return values
 
     def _check_analytic_dimension(self):
-        dimension_ids = self.mapped('analytic_dimension_id').ids
-        if len(self.ids) != len(dimension_ids):
+        tags_with_dimension = self.filtered(
```
pedrobaeza: The same about lambda

analytic_tag_dimension/views/analytic_view.xml
```diff
@@ -11,7 +11,7 @@
                         <field name="code"/>
                     </group>
                     <group string="Tags">
-                        <field name="analytic_tag_ids" nolabel="1"/>
+                        <field name="analytic_tag_ids" nolabel="1" readonly="1"/>
```
pedrobaeza: Don't you put required in view?

pedrobaeza (changes requested):

analytic_tag_dimension/views/analytic_view.xml
angelmoya: No, finally I think that you can create other tags without dimension

angelmoya: Other changes done

pedrobaeza: Why readonly?

Cedric-Pigeon (approved):
Code review. Thanks

analytic_tag_dimension/views/analytic_view.xml
angelmoya: Thanks @pedrobaeza , I was wrong because I was thinking only in case that you already have analytic tags, and you want prevent to create new one on dimension form

angelmoya:
@pedrobaeza, changes done

pedrobaeza (approved):

Cedric-Pigeon (approved):

Cedric-Pigeon:
@pedrobaeza Does anyone can merge this one?  Other PR's are red because of it. Thanks

pedrobaeza:
Yeah, merging. But we have already saturated Travis, hehe...

PR #99, opened 2017-05-12
[IMP] account_analytic_parent: add tree view to adviser wizard

cubells:
This improve tree view in analytic account wizard for advisers.

The tree view now is like this:

6ed3645e-2b41-11e7-8974-54b3da361d81

cc @Tecnativa @jbeficent

### DIFF
--- a/account_analytic_parent/__init__.py
+++ b/account_analytic_parent/__init__.py
@@ -6,3 +6,4 @@
 # Copyright 2017 Serpent Consulting Services Pvt. Ltd.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 from . import models
+from . import wizard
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Don't you remove the other view?

cubells:
@pedrobaeza :

No. The other view is used in Invoicing>Configuration>Analytic accounting>Analytic accounts for maintenance.

pedrobaeza:
I mean the one created under business intelligence menu.

cubells:
Why do you want to remove that view?

pedrobaeza:
Because it's redundant.

cubells:
Then only removing menu is enough. 
We have old tree view in configuration and new tree view for advisers?
Am I right?

pedrobaeza:
Yeah, that's it

pedrobaeza (approved):
Only one thing more it would be great: create some demo data with the parent accounts and the hierarchy we had on v8

pedrobaeza:
@jbeficent @aheficent please review

pedrobaeza:
I'm remembering another thing: please ensure migration path storing in OpenUpgrade the column and not dropping the view accounts (if not yet done), and here restore them with the appropriate fields.

AaronHForgeFlow (approved):
:+1: I agree. Better this view in the Chart of analytic accounts than in the Business intelligence menu.

PR #102, opened 2017-05-24
[WIP][10.0][MIG] account_asset_analytic

luismontalba:
Account asset analytic
==================
This module allows you to define an analytic account per asset and creates
expenses moves with that analytic account instead of the analytic account
per category, as the standard does.
It also propagates analytic account from invoice line to analytic account line.

cc @Tecnativa


account_asset_analytic/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
```
pedrobaeza: You can remove copyright from init files.

account_asset_analytic/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# © 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
```
pedrobaeza: Replace `©` by `Copyright` and add yours

account_asset_analytic/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Account asset analytic",
+    "summary": "Adds analytic account per asset",
+    "version": "10.0.1.0.0",
+    "category": "Analytic Accounting",
+    "website": "http://www.tecnativa.com",
```
pedrobaeza: https

account_asset_analytic/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "author": "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "account_asset",
```
pedrobaeza, line 18: Doesn't it depend on `analytic`?

account_asset_analytic/models/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+# © 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
```
pedrobaeza: Remove copyright here

pedrobaeza (approved):
Little changes, but overall OK

account_asset_analytic/__manifest__.py
luismontalba: Yes, but throw this route:
account_asset > account_accountant > account > analytic

pedrobaeza (changes requested):
The drop in coverage means that there are certain things untested. See here what it's untested and check, because I think that the analytic account propagation to the asset is not correct:

https://codecov.io/gh/OCA/account-analytic/compare/8e5514369405d8bb2980c65add9b41a1666d0f43...4f61bdaf9d6969b32b0edda3380efadde02af1b6/diff

luismontalba:
Runbot can't build due to other modules test

pedrobaeza:
Runbot is building, but it fails with a warning. Anyway, it's possible to test the module in runbot.

pedrobaeza:
@cubells can you review?

cubells (approved):
:+1: 

Code review and tested

pedrobaeza (approved):

PR #110, opened 2017-06-20
[8.0] analytic distribution in asset

drl9319:
I create a new module for insert analytic distribution in assets like the module account_asset_analytic insert 
an account. 

@pedrobaeza @antespi Hello, I create a new module based in your module account_asset_analytic. 
How do you see this new module?


distribution_asset_analytic/__openerp__.py
```diff
@@ -0,0 +1,23 @@
...
+    "category": "Analytic Accounting",
+    "website": "http://praxya.com/",
+    "author": "Tecnativa, "
+              "Praxya, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "account_asset",
```
pedrobaeza: It should depend on account_analytic_plans

pedrobaeza:
Thanks for the contribution, Daniel. Module name `account_asset_analytic_distribution` should be more appropriate.

As a side note, we have a complete set of modules *_distribution in v9/v10 that replaces old *_plans one.

pedrobaeza (approved):
You should link drodriguez@praxya.com to your GitHub account.

PR #111, opened 2017-06-22
[MIG] procurement_analytic module: migration to 10.0

cubells:
- Update Views
- Update README

cc @Tecnativa


procurement_analytic/README.rst
```diff
@@ -0,0 +1,63 @@
...
+or manufacturing orders.
+
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to your user settings.
+#. Enable *Analytic Accounting for Purchases* in *Technical Settings'.
```
pedrobaeza: This text is not proper, as maybe you don't have Purchase installed.

pedrobaeza (changes requested):
Squash together Carlos's commits

procurement_analytic/README.rst
cubells: I need to remove Configuration and Usage sections because this module per se, without install inventory, purchase or sale modules, It can't be configured.

pedrobaeza: OK then, just put it on the main section that this module is only the base for others.

cubells:
@pedrobaeza done.

pedrobaeza (approved):

procurement_analytic/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    'summary': 'This module adds analytic account to procurements',
+    'version': '10.0.1.0.0',
+    'category': 'Analytic',
+    'license': 'AGPL-3',
+    'author': "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    'website': 'https://www.tecnativa.com',
+    'depends': [
+        'procurement',
+    ],
```
luismontalba: Isn't depending on 'purchase' too? Because of the groups restriction in the view:
groups="purchase.group_analytic_accounting"

cubells:
@luismontalba you are right.

I set correct group and dependencies.

Thanks.

luismontalba (approved):

elicoidal (approved):

PR #114, opened 2017-06-27
[MIG] account_analytic_parent: Migration to 10.0

chienandalu:
Account Analytic Parent
=======================

This module reintroduces the hierarchy to the analytic accounts as
it was in previous versions of Odoo. This module is a base module for
other modules to manage the hierarchy concept in analytics.

cc @Tecnativa


AaronHForgeFlow (approved):
:+1: 
Functional test + code review. Thank you for this contribution!

account_analytic_parent/models/__init__.py
```diff
@@ -1,8 +1,4 @@
 # -*- coding: utf-8 -*-
-# Copyright 2017 Eficent Business and IT Consulting Services S.L.
-# Copyright 2017 Luxim d.o.o.
-# Copyright 2017 Matmoz d.o.o.
-# Copyright 2017 Deneroteam.
-# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
```
LoisRForgeFlow: Why are you removing credits here? is it a new convention?

LoisRForgeFlow: @chienandalu

pedrobaeza: Because `__init__.py` are not code by themselves, and making light these files is better to maintain.

LoisRForgeFlow: Ok, thanks for the explanation @pedrobaeza :wink:

account_analytic_parent/__manifest__.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# Copyright 2017 Luxim d.o.o.
+# Copyright 2017 Matmoz d.o.o.
+# Copyright 2017 Deneroteam.
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# Coypyright 2017 Tecnativa
```
elicoidal: copyright

account_analytic_parent/models/account_analytic_account.py
```diff
@@ -0,0 +1,68 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class AccountAnalyticAccount(models.Model):
+    _inherit = 'account.analytic.account'
+
+    parent_id = fields.Many2one(
```
elicoidal, line 16: Should not we add the type (View, Regular) and if View type can be used as parent and not possible to post entries

elicoidal (changes requested):
Functionally I would suggest to reintroduce the type as well

account_analytic_parent/models/account_analytic_account.py
chienandalu: Thanks for your kind comments @elicoidal. Isn't `active = False` (new to v10 ) enough to get the same functionality that `type = view` gave.

pedrobaeza: The problem here is that you will need then to add glue modules for each module (sale, purchase) to filter the domain of the analytic accounts.

The way to achieve the same should be:

- If the parent account is of type 'normal', you let it active.
- If the parent account is of type 'view', you deactivate it after create it.

account_analytic_parent/models/account_analytic_account.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @api.multi
+    def _compute_debit_credit_balance(self):
+        """
+        Warning, this method overwrites the standard because the hierarchy
+        of analytic account changes
+        """
+        super(AccountAnalyticAccount, self)._compute_debit_credit_balance()
+        for account in self:
+            debit = account.debit
```
pedrobaeza: Reduce this to:
```
account.debit += sum(account.mapped('child_ids.debit'))
...
```

account_analytic_parent/models/account_analytic_account.py
```diff
@@ -0,0 +1,68 @@
...
+    def check_recursion(self):
+        for account in self:
+            if not super(AccountAnalyticAccount, account)._check_recursion():
+                raise UserError(
+                    _('You can not create recursive analytic accounts.'),
+                )
+
+    @api.multi
+    @api.onchange('parent_id')
+    def on_change_parent(self):
```
pedrobaeza: Call it the standard way: `_onchange_parent_id`

account_analytic_parent/models/account_analytic_account.py
```diff
@@ -0,0 +1,68 @@
...
+        for account in self:
+            account.partner_id = account.parent_id.partner_id or False
```
pedrobaeza: No need for the `False` part

pedrobaeza (changes requested):

account_analytic_parent/models/account_analytic_account.py
elicoidal: @pedrobaeza maybe I dont understand well your proposal but it does not fit well usabiilty:
Inactive means "Do not use" and brings confusion in your design.
Besides how do you display the View accounts if they are inactive and how you differentiate them from the actual inactive ones?

Having the type in the base modules looks cleaner and standard feature (same as in v8): later you indeed have to add domain control in the depending modules but I don't understand why you would need glue modules.

chienandalu:
@pedrobaeza Changes done

account_analytic_parent/models/account_analytic_account.py
pedrobaeza: It's not a question of usability. It's a question of availability. You don't have available on each of the places where analytic accounts are displayed (sale, purchase, invoice, moves...) the domain for filtering out "view" accounts unless you depend on all of them, so this other field makes the trick, although it mixes 2 concepts, but in some way related, as the goal is the same: to not allow to select that accounts. Anyway, there's a possible way of differentiating them: deactivated account with children, it's a view account. If not, it's a deactivated account.

I put an example of the glue modules you will need it:
```
from odoo import fields, models

class PurchaseOrderLine(models.Model):
    _inherit = 'purchase.order.line'

    analytic_account_id = fields.Many2one(domain="[('type', '!=', 'view')]")
```

pedrobaeza (approved):

account_analytic_parent/models/account_analytic_account.py
elicoidal: ok

elicoidal (approved):

PR #116, opened 2017-06-27
[ADD] sale_procurement_analytic: Migration to 10.0

cubells:
Module to propagate analytic account from sale order to procurement.

cc @Tecnativa


sale_procurement_analytic/models/product.py
```diff
@@ -0,0 +1,18 @@
...
+
+from odoo import api, models
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.multi
+    def _need_procurement(self):
+        # When sale/product is installed alone, there is no need to create
```
AaronHForgeFlow: This is done the standard Odoo, isn't it?

AaronHForgeFlow (commented):
Thank you for this module. Functional test in local was fine. Runbot is not working because the dependency to procurement_analytic. Can you please rebase your branch?

sale_procurement_analytic/models/product.py
pedrobaeza: Yeah, this is something outside the scope of this module. Remove it.

sale_procurement_analytic/models/sale.py
```diff
@@ -0,0 +1,17 @@
...
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    @api.multi
+    def _prepare_order_line_procurement(self, group_id=False):
+        self.ensure_one()
+        res = super(SaleOrderLine, self)._prepare_order_line_procurement(
+            group_id=group_id)
+        res['account_analytic_id'] = self.order_id.related_project_id.id
```
pedrobaeza: Better to use `project_id` directly. By the way, what a bad name! It's the analytic account, not the project. Please put a comment indicating that the field contains actually the analytic account, not the project, so that anyone that sees this code doesn't wonder the same

sale_procurement_analytic/tests/test_sale_procurement_analytic.py
```diff
@@ -0,0 +1,43 @@
...
+from odoo.tests import common
+
+
+class TestSaleProcurementAnalytic(common.SavepointCase):
+    """ Use case : Prepare some data for current test case """
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestSaleProcurementAnalytic, cls).setUpClass()
+
```
pedrobaeza: Remove this extra line

sale_procurement_analytic/tests/test_sale_procurement_analytic.py
```diff
@@ -0,0 +1,43 @@
...
+                'product_uom_qty': 1,
+                'price_unit': cls.product.list_price,
+                'name': cls.product.name,
+            })],
+        })
+
+    def test_sale_to_procurement(self):
+        # Confirm the sale order
+        self.sale_order.action_confirm()
+
```
pedrobaeza: Remove this extra line

pedrobaeza (changes requested):

sale_procurement_analytic/models/product.py
cubells: @aheficent No, procurement is not created if only sale/product is installed.

@pedrobaeza  I think it's necessary.

This module has **only** sale and procurement_analytic dependencies.

Without this method, procurement is not ever created at all. Tests fail too. 
If others modules are installed, then procurement is created as you expected.

pedrobaeza: Yeah, I understand you, but this will create procurement for everything, which is not correct. This will only create procurements on MTO/dropshipping products. Make one thing, add the condition that this should be used only when testing the module (something similar to https://github.com/OCA/bank-payment/blob/8.0/account_payment_transfer_reconcile_batch/models/payment_order.py#L28 - with one condition less as you don't need to call the same method 2 times).

sale_procurement_analytic/models/sale.py
cubells: @pedrobaeza done!

pedrobaeza (approved):
@luismontalba please review

luismontalba:
Same problem in runbot when trying to install the module:
"You try to install module 'sale_procurement_analytic' that depends on module 'procurement_analytic'.
But the latter module is not available in your system". I think same as @aheficent , may be a rebase is needed.

cubells:
I close and reopen a new pr to be able to build runbot.

PR #118, opened 2017-07-11
[10.0][FIX] account_asset_analytic: Fix create_move

luismontalba:
cc @Tecnativa

### DIFF
--- a/account_asset_analytic/models/account_asset_depreciation_line.py
+++ b/account_asset_analytic/models/account_asset_depreciation_line.py
@@ -9,8 +9,9 @@ class AccountAssetDepreciationLine(models.Model):
     _inherit = 'account.asset.depreciation.line'
 
     @api.multi
-    def create_move(self):
-        move_ids = super(AccountAssetDepreciationLine, self).create_move()
+    def create_move(self, post_move=True):
+        move_ids = super(AccountAssetDepreciationLine, self).create_move(
+            post_move=post_move)
         for line in self:
             asset = line.asset_id
             analytic = asset.analytic_account_id
--- a/account_asset_analytic/__manifest__.py
+++ b/account_asset_analytic/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Account asset analytic",
     "summary": "Adds analytic account per asset",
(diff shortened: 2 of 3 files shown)


account_asset_analytic/models/account_asset_depreciation_line.py
```diff
@@ -9,7 +9,7 @@ class AccountAssetDepreciationLine(models.Model):
     _inherit = 'account.asset.depreciation.line'
 
     @api.multi
-    def create_move(self):
+    def create_move(self, post_move=True):
         move_ids = super(AccountAssetDepreciationLine, self).create_move()
```
pedrobaeza: You are not passing the same argument to super

luismontalba: Fixed and tested (new commit squashed)

account_asset_analytic/models/account_asset_depreciation_line.py
```diff
@@ -9,8 +9,9 @@ class AccountAssetDepreciationLine(models.Model):
     _inherit = 'account.asset.depreciation.line'
 
     @api.multi
-    def create_move(self):
-        move_ids = super(AccountAssetDepreciationLine, self).create_move()
+    def create_move(self, post_move=True):
+        move_ids = super(AccountAssetDepreciationLine, self).create_move(
+            post_move=True)
```
pedrobaeza: This should be `post_move=post_move` (the value received by argument). One thing is the default value (True), which is used in the signature in case this argument is not passed, and the other is the argument passed to super, which is the value contained on `post_move` (which can be the default or not, but that doesn't matter you).

luismontalba: Re-fixed :smile:

pedrobaeza:
One last change: always remember to change version number according this rule: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

luismontalba:
Version number changed

pedrobaeza (approved):

pedrobaeza:
Fast-tracking when CIs finish as this makes Odoo to crash.

PR #128, opened 2017-08-24
[10.0][ADD] module account_analytic_default_tag

grindtildeath:
Account Analytic Default Tag
============================

This module extends the functionality of account_analytic_default to support
analytic tags and to allow you to set different tags as default value.


analytic_tag_default/views/account_analytic_default.xml
```diff
@@ -0,0 +1,28 @@
...
+    <record id="view_account_analytic_default_form" model="ir.ui.view">
+        <field name="name">account.analytic.default.form</field>
+        <field name="model">account.analytic.default</field>
+        <field name="inherit_id" ref="account_analytic_default.view_account_analytic_default_form" />
+        <field name="arch" type="xml">
+            <field name="analytic_id" position="attributes">
+                <attribute name="required">0</attribute>
+            </field>
+            <field name="analytic_id" position="after">
+                <field name="tag_ids" widget="many2many_tags" groups="analytic.group_analytic_accounting" />
```
jcoux: Can you move the groups restriction on the definition of this inherited view?

With that, the `analytic_id` field keep the required flag if we don't have rights on tags.

analytic_tag_default/models/account_analytic_default.py
```diff
@@ -0,0 +1,36 @@
...
+class AccountInvoiceLine(models.Model):
+    _inherit = "account.invoice.line"
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
+        res = super(AccountInvoiceLine, self)._onchange_product_id()
+        rec = self.env['account.analytic.default'].account_get(
+            self.product_id.id, self.invoice_id.partner_id.id, self.env.uid,
+            fields.Date.today(), company_id=self.company_id.id)
+        self.analytic_tag_ids = rec.tag_ids.ids
```
jcoux: I think if we don't have rights on tag, we don't want to change `analytic_tag_ids` value.

analytic_tag_default/models/account_analytic_default.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @api.one
+    @api.constrains('analytic_id', 'tag_ids')
+    def _check_required(self):
+        # As at least one field of analytic_id and tag_ids is required we throw
+        # a validation error should both fields be empty
+        if not self.analytic_id and not self.tag_ids:
+            raise ValidationError(_(
+                'At least one Analytic Account or Analytic Tag has to be '
+                'defined.'))
```
jcoux: If you move the groups restriction on the definition of the inherited view (see another comment), you don't need to do this check if user don't have rights on tags.

jcoux (changes requested):

analytic_tag_default/models/account_analytic_default.py
grindtildeath: This behaviour is the same that the one from odoo module account_analytic_default : https://github.com/odoo/odoo/blob/10.0/addons/account_analytic_default/models/account_analytic_default.py#L60
IMO we should keep it the same.

analytic_tag_default/views/account_analytic_default.xml
grindtildeath: In the original inherited view, the group is already defined on analytic_id :
https://github.com/odoo/odoo/blob/10.0/addons/account_analytic_default/views/account_analytic_default_view.xml#L26
So if this module is installed, there is no way that you have rights on analytic_id but not on tag_ids.

jcoux: I have not see that the group is the same on the 2 fields.
But in this case, if you don't have the group, you never can create an `account.analytic.default` object with your constrains, because you can't define one of this 2 fields.

analytic_tag_default/models/account_analytic_default.py
jcoux: Even if odoo module do the onchange like that, I think yet that if you don't have rights to see a field, you must not have the rights to change this same field.

grindtildeath: Well, I just made a few tests again using two different users and we definitely need this behaviour.
Consider user A (part of group_analytic_accounting) created an account.analytic.default to add analytic account ABC on product XYZ.
Now if user B (part of accounting group only) creates an account.invoice with product XYZ, we want the account.invoice.line to have analytic account ABC, even if this user is part of group_analytic_accounting. By the way this is the behaviour from odoo.

So as we now add the possibility to add analytic.tag on account.analytic.default, we need this update because these tags shall be applied with the same conditions as analytic.account.

analytic_tag_default/views/account_analytic_default.xml
grindtildeath: Fact is I couldn't access these views with a user not part of group_analytic_accounting.
My logic was to keep everything as close as possible to what's in odoo's code.
Maybe someone from the community can give an input ?

grindtildeath:
No idea why the tests are failing here as I have no issue on my local instance using either pytest or odoo --test-enable.

grindtildeath:
@jcoux  @yvaucher @TDu Can you please pass by for a small review ?

account_analytic_default_tag/models/account_analytic_default.py
```diff
@@ -0,0 +1,46 @@
...
+                raise ValidationError(_(
+                    'At least one Analytic Account or Analytic Tag has to be '
+                    'defined.'))
+
+    @api.model
+    def create(self, vals):
+        if 'analytic_id' not in vals and 'analytic_tag_ids' not in vals:
+            raise ValidationError(_(
+                'At least one Analytic Account or Analytic Tag has to be '
+                'defined.'))
```
jcoux, line 32: With constraints you defined `_check_required`, do we need to do the check again here?

grindtildeath: IIRC the check with api.constrains, won't happen if those fields are not used in the record creation, but I should probably have added a comment to explicit this.

jcoux: Thanks for explanations :)

jcoux (approved):
Code review only.
LGTM :+1:

PR #112, opened 2017-06-26
[MIG] purchase_procurement_analytic module: migration to 10.0

cubells:
Depends on
- [x] #111  

Migration of the module purchase_procurement_analytic:

- Update views
- Update README
- Update tests

cc @Tecnativa

PR #131, opened 2017-09-14
[MIG] purchase_procurement_analytic module: migration to 10.0

cubells:
Related: https://github.com/OCA/account-analytic/pull/112

Migration of the module purchase_procurement_analytic:

- Update views
- Update README
- Update tests

cc @Tecnativa


pedrobaeza (approved):

pedrobaeza:
Merging as this has been previously proposed.

purchase_procurement_analytic/models/stock.py
```diff
@@ -0,0 +1,18 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import api, models
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    @api.model
+    def _prepare_procurement_from_move(self):
```
rousseldenis, line 13: @cubells It seems that this piece of code should have been placed in a more generic module 'stock_procurement_analytic' for example (it does not imply purchases necessarily...).

What do you think about it ?

PR #152, opened 2018-01-19
[10.0][ADD] Add default analytic account for accounts

p-tombez:
Add the ability to set default analytic account on accounts.
This is useful when, e.g. you use the account_analytic_required
and have multi-currencies, thus need to create an automatic
currency exchange difference entry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Available addons
 ----------------
 addon | version | summary
 --- | --- | ---
+[account_analytic_default_account](account_analytic_default_account/) | 10.0.1.0.0 | Add default analytic account for accounts
 [account_analytic_distribution](account_analytic_distribution/) | 10.0.1.0.0 | Distribute incoming/outcoming account moves to several analytic accounts
 [account_analytic_distribution_required](account_analytic_distribution_required/) | 10.0.1.0.0 | Account Analytic Distribution Required
 [account_analytic_no_lines](account_analytic_no_lines/) | 10.0.1.0.0 | Hide analytics lines and disable their generation from a move line.
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ Available addons
 ----------------
 addon | version | summary
 --- | --- | ---
-[account_analytic_default_account](account_analytic_default_account/) | 10.0.1.0.0 | Add default analytic account for accounts
 [account_analytic_distribution](account_analytic_distribution/) | 10.0.1.0.0 | Distribute incoming/outcoming account moves to several analytic accounts
 [account_analytic_distribution_required](account_analytic_distribution_required/) | 10.0.1.0.0 | Account Analytic Distribution Required
 [account_analytic_no_lines](account_analytic_no_lines/) | 10.0.1.0.0 | Hide analytics lines and disable their generation from a move line.
--- a/account_analytic_default_account/README.rst
+++ b/account_analytic_default_account/README.rst
@@ -12,8 +12,7 @@ accounts.
 Configuration
 =============
 
-For example, if you want to have an analytic account whenever you use
-a given account, then go to *Accounting -> Configuration -> Analytic Defaults*
+Go to *Accounting -> Configuration -> Analytic Defaults*
 and set the corresponding settings.
 
 Usage
--- a/account_analytic_default_account/README.rst
+++ b/account_analytic_default_account/README.rst
@@ -1,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License
+
+================================
+Account Analytic Default Account
+================================
+
+This module adds the ability to set a default analytic account on
+accounts.
+
+Configuration
+=============
+
+For example, if you want to have an analytic account whenever you use
+a given account, then go to *Accounting -> Configuration -> Analytic Defaults*
+and set the corresponding settings.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/87/10.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/account-analytic/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Patrick Tombez <patrick.tombez@camptocamp.com>
+
+Maintainer
+----------
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/account_analytic_default_account/models/account_analytic_default_account.py
+++ b/account_analytic_default_account/models/account_analytic_default_account.py
@@ -1,0 +1,130 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class AccountAnalyticDefaultAccount(models.Model):
+    _inherit = "account.analytic.default"
+
+    account_id = fields.Many2one(
+        'account.account', string='Account',
+        ondelete='cascade',
+        help="Select an account which will use analytic account specified in "
+             "analytic default (e.g. create new customer invoice or "
+             "Sales order if we select this product, "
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -18,6 +18,7 @@ Available addons
 ----------------
 addon | version | summary
 --- | --- | ---
+[account_analytic_default_account](account_analytic_default_account/) | 10.0.1.0.0 | Add default analytic account for accounts
```
simahawk: skip it: this part is updated automatically ;)

account_analytic_default_account/README.rst
```diff
@@ -0,0 +1,59 @@
...
+Account Analytic Default Account
+================================
+
+This module adds the ability to set a default analytic account on
+accounts.
+
+Configuration
+=============
+
+For example, if you want to have an analytic account whenever you use
```
simahawk: skip the 1st part. Just state "Go to Accounting..."

account_analytic_default_account/models/account_analytic_default_account.py
```diff
@@ -0,0 +1,130 @@
...
+from odoo import api, fields, models
+
+
+class AccountAnalyticDefaultAccount(models.Model):
+    _inherit = "account.analytic.default"
+
+    account_id = fields.Many2one(
+        'account.account', string='Account',
+        ondelete='cascade',
+        help="Select an account which will use analytic account specified in "
```
simahawk: can you rephrase this? The way I get is: "This account will be used by default for sale orders or invoices" or something similar :)

account_analytic_default_account/models/account_analytic_default_account.py
```diff
@@ -0,0 +1,130 @@
...
+             "analytic default (e.g. create new customer invoice or "
+             "Sales order if we select this product, "
+             "it will automatically take this as an analytic account)"
+    )
+
+    @api.model
+    def account_get(self, product_id=None, partner_id=None, user_id=None,
```
simahawk: I'd split this in more methods and simplify it a bit, for instance:
```python
    account_get_domain_keys = (
        'product_id', 'partner_id', 'user_id',
        'date', 'company_id', 'account_id',
    )

    def _account_get_domain(self, **kw):
        """Build account.analytic.default domain.

        :param kw: filter keys.
            Available filter keys are defined into `account_get_domain_keys`.
        :return: a search domain with all required keys' leaves.
            Each key will be searched for both False and specified value.
            For instance: [
                ('product_id'), '=', False), '|', [('product_id'), '=', 100)
            ]
        """
        domain = []
        account_get_domain_keys = self.account_get_domain_keys[:]
        # date will be handled differently
        date_key = account_get_domain_keys.pop('date', None)

        for key in account_get_domain_keys:
            if kw.get(key):
                domain += ['|', (key, '=', kw.get(key))]
            domain += [(key, '=', False)]

        if date_key and kw.get(date_key):
            domain += ['|', ('date_start', '<=', kw.get(date_key)),
                       ('date_start', '=', False)]
            domain += ['|', ('date_stop', '>=', kw.get(date_key)),
                       ('date_stop', '=', False)]
        return domain

    def account_get(self, **kw):
        domain = self._account_get_domain(**kw)
        best_index = -1
        res = self.env['account.analytic.default']
        for rec in self.search(domain):
            index = 0
            for k in kw.keys():
                if rec[k]:
                    index += 1
            if index > best_index:
                res = rec
                best_index = index
        return res
```

account_analytic_default_account/models/account_analytic_default_account.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class AccountInvoiceLine(models.Model):
+    _inherit = "account.invoice.line"
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
+        res = super(AccountInvoiceLine, self)._onchange_product_id()
+        rec = self.env['account.analytic.default'].account_get(
+            self.product_id.id, self.invoice_id.partner_id.id, self.env.uid,
```
simahawk: the method is supposed to accept keyword args

account_analytic_default_account/models/account_analytic_default_account.py
```diff
@@ -0,0 +1,130 @@
...
+            fields.Date.today(), company_id=self.company_id.id,
+            account_id=self.account_id.id)
+        self.account_analytic_id = rec.analytic_id.id
+        return res
+
+    def _set_additional_fields(self, invoice):
+        if not self.account_analytic_id:
+            rec = self.env['account.analytic.default'].account_get(
+                self.product_id.id, self.invoice_id.partner_id.id,
```
simahawk: same

simahawk (changes requested):

simahawk:
would you mind adding some tests for this?

account_analytic_default_account/models/account_analytic_default_account.py
p-tombez: I can't change the method signature so drastically as it already exists in the Odoo addons. I can't guarantee that the calls are made with kwargs everywhere. That's the main reason I copied the method and updated it for my needs.

simahawk: fine, you are right. Keep the signature as it is... but respect it in your calls if possible :)
Then you can still address the rest I guess ;)

p-tombez:
@simahawk Thanks for the suggestions! I also added some tests

account_analytic_default_account/models/account_analytic_default_account.py
```diff
@@ -0,0 +1,145 @@
...
+        ondelete='cascade',
+        help="Select the account corresponding to an analytic account "
+             "which will be used on the lines of invoices or account moves"
+    )
+
+    def _account_get_domain_keys(self):
+        """Returns a set of original account_get arg names - self + new ones"""
+        func = super(AccountAnalyticDefaultAccount, self).account_get
+        return (
+            list(func.func_code.co_varnames[1:func.func_code.co_argcount]) +
```
simahawk: ???

account_analytic_default_account/models/account_analytic_default_account.py
```diff
@@ -0,0 +1,145 @@
...
+
+        date_val = kw.get('date')
+        if date_val:
+            domain += ['|', ('date_start', '<=', date_val),
+                       ('date_start', '=', False)]
+            domain += ['|', ('date_stop', '>=', date_val),
+                       ('date_stop', '=', False)]
+        return domain
+
+    def _account_get(self, **kw):
```
simahawk: docstring explaining how you get to the right account

account_analytic_default_account/models/account_analytic_default_account.py
```diff
@@ -0,0 +1,145 @@
...
+                    index += 1
+            if index > best_index:
+                res = rec
+                best_index = index
+        return res
+
+    @api.model
+    def account_get(self, *args, **kwargs):
+        """Wrapper method to convert *args to **kwargs"""
+        kwargs.update(zip(self._account_get_domain_keys(), args))
```
simahawk: don't use black magic. If the original method has some specific keys, use them.

simahawk (changes requested):

account_analytic_default_account/models/account_analytic_default_account.py
```diff
@@ -0,0 +1,148 @@
...
+                    index += 1
+            if index > best_index:
+                res = rec
+                best_index = index
+        return res
+
+    @api.model
+    def account_get(self, product_id=None, partner_id=None, user_id=None,
+                    date=None, company_id=None, account_id=None):
+        """Wrapper method to convert positional args to kwargs"""
```
simahawk: you don't need any wrapping:
```
@api.model
def account_get(self, product_id=None, partner_id=None, user_id=None,
                date=None, company_id=None, account_id=None):

    filters = {
        'product_id': product_id,
        'partner_id': partner_id,
        'user_id': user_id,
        'date': date,
        'company_id': company_id,
        'account_id': account_id,
    }
    domain = self._account_get_domain(**filters)
    keys = filters.keys()
    [...]
```

account_analytic_default_account/models/account_analytic_default_account.py
```diff
@@ -0,0 +1,148 @@
...
+        :param kw: filter keys.
+            Available filter keys are defined into `account_get_domain_keys`.
+        :return: a search domain with all required keys' leaves.
+            Each key will be searched for both False and specified value.
+            For instance: [
+                ('product_id'), '=', False), '|', [('product_id'), '=', 100)
+            ]
+        """
+        domain = []
+        account_get_domain_keys = self.account_get_domain_keys[:]
```
simahawk: you don't need this anymore as the keys are controlled by the main method signature
```
def _account_get_domain(self, **kw):
    """Build account.analytic.default domain.
    :param kw: filter keys.
        Available filter keys are defined into `account_get_domain_keys`.
    :return: a search domain with all required keys' leaves.
        Each key will be searched for both False and specified value.
        For instance: [
            ('product_id'), '=', False), '|', [('product_id'), '=', 100)
        ]
    """
    domain = []
    date_val = kw.pop('date', None)
    for key, value in kw.items():
        if value is not None:
            domain += ['|', (key, '=', value)]
        domain += [(key, '=', False)]

    if date_val:
        domain += ['|', ('date_start', '<=', date_val),
                   ('date_start', '=', False)]
        domain += ['|', ('date_stop', '>=', date_val),
                   ('date_stop', '=', False)]
    return domain
```

simahawk (changes requested):

simahawk (approved):

simahawk:
remember to squash before merge

leemannd (approved):

fclementic2c:
cannot test it - runbot does not initate as normally :\

feketemihai (approved):

feketemihai:
@pedrobaeza can you please metge this, i tested on tunbot as well..

PR #146, opened 2017-12-28
[11.0][MIG] analytic_tag_dimension

ghost:

PR #153, opened 2018-01-30
[FIX] analytic_tag_dimension: fixed grouping in reports->invoices

mpanarin:
fixes the same bug that was found in v11:
https://github.com/OCA/account-analytic/pull/146#issuecomment-360768397


yvaucher (approved):
LGTM

yvaucher:
fast forwarded as merged in version 11.0

PR #155, opened 2018-02-13
[8.0] [MIG] purchase analytic distribution

PieterPaulussen:
Depends on https://github.com/OCA/account-analytic/pull/156


StefanRijnhart:
Only when #156 is merged we can check Travis outcome but the code looks good to me.

PR #156, opened 2018-02-13
[8.0] [MIG] account analytic distribution

PieterPaulussen:


account_analytic_distribution/__openerp__.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Antonio Espinosa - <antonio.espinosa@tecnativa.com>
+# Copyright 2017 Vicent Cubells - <vicent.cubells@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
```
StefanRijnhart, line 6: The travis lintchecker breaks on missing 'license' key in this file.

StefanRijnhart:
Tests break on the combination with other modules in this repo that depend on account_analytic_plans. Can you try to partition the tests using INCLUDE/EXCLUDE as in https://github.com/OCA/sale-workflow/blob/8.0/.travis.yml#L26? You can just set this module apart.

StefanRijnhart:
~~Remaining failure is in another module.~~

account_analytic_distribution/models/account_move_line.py
```diff
@@ -0,0 +1,33 @@
...
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    analytic_distribution_id = fields.Many2one(
+        comodel_name='account.analytic.distribution',
+        string='Analytic distribution',
+    )
+
+    def _analytic_line_distributed_prepare(self, line, rule):
+        res = self._prepare_analytic_line(line)
```
StefanRijnhart: I'm guessing you changed this method from an implicit api.multi to an implicit api.model because this is what happened in Odoo itself with _prepare_analytic_line. I have a slight preference to minimize this change as follows.
```
@api.multi
def _analytic_line_distributed_prepare(self, rule):
        res = self._prepare_analytic_line(self)
```

StefanRijnhart: Pushing this change myself @PieterPaulussen because if nothing else it fixes the last travis error: a line length error precisely because of the signature change.

StefanRijnhart (approved):
Code review: checked the diff with the 9.0 version.

PR #159, opened 2018-02-21
[10.0] Add module product_analytic_sale

alexis-via:
With this PR, we now have a full set : product_analytic + product_analytic_pos + product_analytic_purchase

### DIFF
--- a/product_analytic_sale/README.rst
+++ b/product_analytic_sale/README.rst
@@ -1,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+Product Analytic Sale
+=====================
+
+This module is a glue module between the *product_analytic* and
+*sale* modules.
+
+Usage
+=====
+
+When you create an invoice from a sale order with products that have an
+**expense analytic account** (or whose internal category has an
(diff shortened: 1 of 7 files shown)


product_analytic_sale/README.rst
```diff
@@ -0,0 +1,58 @@
...
+=====================
+
+This module is a glue module between the *product_analytic* and
+*sale* modules.
+
+Usage
+=====
+
+When you create an invoice from a sale order with products that have an
+**expense analytic account** (or whose internal category has an
```
feketemihai, line 16: should be income analytic account

product_analytic_sale/README.rst
```diff
@@ -0,0 +1,58 @@
...
+**expense analytic account**), it will be automatically set on the
```
feketemihai, line 17: the same, income analytic account

feketemihai (changes requested):
@alexis-via I don't get it why do tou need this module extensions, since you only use them for invoice lines, and you already have in product_analytic https://github.com/OCA/account-analytic/blob/10.0/product_analytic/models/account_invoice.py#L32, why do you need the prepare for purchase, pos, sale since on create it's already checking if no analytic, will be added depending on the type of invoice...

alexis-via:
@feketemihai Yes, you are probably right... this module product_analytic_sale is not needed.
The module product_analytic_pos is required because pos doesn't create invoices. The module product_analytic_purchase sets the analytic on the purchase order line also.

PR #138, opened 2017-11-09
[FIX][10.0]Pass the analytic account from the procurement to the move

AaronHForgeFlow:
It solves issue: https://github.com/OCA/account-analytic/issues/137

Current behaviour: When a procurement has an analytic account ii will not pass the analytic information to the stock move. It has to be set manually.

Expected behaviour: The analytic account is passed from the procurement to the stock move.

As the stock module depends on procurement module, the stock analytic module should depend on procurement_analytic module and pass the information from the procurement to the stock move.

PR #161, opened 2018-03-18
[11.0][MIG] stock_analytic

jaredkipe:
Tests passed with minor changes (11.0 specific needs).
I currently use this in something that makes stock moves on its own, and it works perfectly fine, however...

There is a UX problem in Odoo 11.0 as there is no good way to modify the stock move created on a picking to have an analytic account.  I think I should add that to the current tree view (on the picking) to the left of the 'Initial Demand' field.


pedrobaeza:
There's a Travis error

AaronHForgeFlow (commented):
IMHO this module should pass also the analytic account from the procurement to the stock move as it is proposed here. https://github.com/OCA/account-analytic/pull/138

pedrobaeza:
On v11 there's no more procurements...

AaronHForgeFlow:
@pedrobaeza yes, sorry, I thought this was 9.0 :face_with_head_bandage:

AaronHForgeFlow:
Commit history is missing thought

pedrobaeza:
Indeed. Thanks for pointing it. @jaredkipe you should preserve commit history following technical method in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

AaronHForgeFlow (commented):
Probably not due this module but the analytic account is not passed from the purchase line to the stock move. As the analytic account cannot be set manually in the stock move, cannot do functional tests. BTW IMHO it'd be great to show the analytic account in the sock move from the picking form view.

jaredkipe:
Closing for branch in #162

PR #162, opened 2018-03-19
[11.0][MIG] stock_analytic 

jaredkipe:
Tests passed with minor changes (11.0 specific needs).
I currently use this in something that makes stock moves on its own, and it works perfectly fine, however...

There is a UX problem in Odoo 11.0 as there is no good way to modify the stock move created on a picking to have an analytic account. I think I should add that to the current tree view (on the picking) to the left of the 'Initial Demand' field.

( better #161 )


jaredkipe:
I'll be adding a view to expose/pick the analytic account on the picking's move line tree in the near future (probably today).

AaronHForgeFlow (approved):
Tested on local LGTM :+1:

LoisRForgeFlow (approved):
LGTM

AaronHForgeFlow:
@jaredkipe can you squash a bit your commits? Thanks!

CasVissers-360ERP (approved):
LGTM

CasVissers-360ERP:
@jaredkipe 
`        for num in range(0, 2):
            if res[num][2]["account_id"] != self.product_id.\
                    categ_id.property_stock_valuation_account_id.id:
                res[num][2].update({
                    'analytic_account_id': self.analytic_account_id.id,
                })
        return res
`

This function doesn't seem to work. I get an analytic account in both the debit & credit move line. Should be only in the debit line.

EDIT:
@jaredkipe Seems like I configured some settings wrongly. Working now. Functional test ok.

AaronHForgeFlow:
Hi @jaredkipe this is about to be merged. Can you squash a bit some of the commits with similar content(flake8, transifex, etc)?

jaredkipe:
@aheficent All done!

PR #163, opened 2018-03-25
[8.0.1.0.1] stock_analytic

andhit-r:
* Adjust _prepare_account_move_line method with no context param


AaronHForgeFlow (commented):
The original method uses the context, doesn't it? https://github.com/odoo/odoo/blob/8.0/addons/stock_account/stock_account.py

PR #164, opened 2018-03-29
[10.0][ADD] analytic_account_sequence

AaronHForgeFlow:
Generate an automatic and unique code for the analytic hierarchy


analytic_account_sequence/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    'name': 'Analytic account code sequence',
+    'summary': 'Analytic account code sequence',
+    'version': '10.0.1.0.0',
+    'author':   'Eficent, '
+                'SerpentCS ,'
+                'Project Expert Team ,'
+                'Odoo Community Association (OCA)',
+    'website': 'https://www.github.com/OCA/account-analytic',
+    'category': 'Project Management',
+    'depends': ['analytic', 'project', 'account_analytic_parent'],
```
pedrobaeza: project?

analytic_account_sequence/i18n/analytic_account_sequence.pot
```diff
@@ -0,0 +1,266 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove this

analytic_account_sequence/__manifest__.py
AaronHForgeFlow: It depends but it should not, I'll remove all the dependent code.

feketemihai:
@aheficent @pedrobaeza the module is merged in 10 and much simplified, since it uses ir. sequence... I think much simpler to maintain it.. I think this one should be closed..

pedrobaeza:
Yeah, I agree. I have written in the other PR because I remembered that @aheficent was working in something similar.

PR #166, opened 2018-04-02
[10.0] [BUG] Remove accents.

feketemihai:
@pedrobaeza It seems that we still have some issues in comments with french accents.


pedrobaeza (approved):
OK, let's see if this one is the definitive.

PR #168, opened 2018-04-09
11.0 mig account analytic parent

jaumeplanas:


account_analytic_parent/models/account_analytic_account.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @api.multi
+    def _compute_debit_credit_balance(self):
+        """
+        Warning, this method overwrites the standard because the hierarchy
+        of analytic account changes
+        """
+        super(AccountAnalyticAccount, self)._compute_debit_credit_balance()
+        for account in self:
+            account.debit += sum(account.mapped("child_ids.debit"))
```
jcdrubay, line 32: For these 3 lines, I think it should be a `=` instead of a `+=`.

`account.debit = sum(account.mapped("child_ids.debit"))`

jcdrubay (commented):
All seems good :+1: just one tiny adjustments maybe.

account_analytic_parent/wizard/account_analytic_wizard.py
```diff
@@ -0,0 +1,35 @@
...
+
+class AccountAnalyticChart(models.TransientModel):
+    _name = "account.analytic.chart"
+    _description = "Account Analytic Chart"
+
+    from_date = fields.Date(string="From")
+    to_date = fields.Date(string="To")
+
+    @api.multi
+    def analytic_account_chart_open_window(self):
```
jcdrubay: Is this still maintained?
When testing, this wizard opened a form view to create a new Analytic Account. I don't think it is the expected behavior.

I was expecting to get a list of analytic accounts with the debit,credit,balance for the period selected in the wizard.

account_analytic_parent/tests/test_account_analytic_account.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# Copyright 2017 Luxim d.o.o.
+# Copyright 2017 Matmoz d.o.o.
+# Copyright 2017 Deneroteam.
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# Copyright 2017 Tecnativa
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo.tests.common import TransactionCase
+from odoo.exceptions import ValidationError
```
hieulucky111, line 9: import order.

account_analytic_parent/models/account_analytic_account.py
j-zaballa: I think the code is ok as it is. The call to super computes debit, credit and balance for every account. Then the method adds debit, credit and balance from child accounts to parent. If we use `=` instead of `+=`, we lose the information of the account itself and only get the sum from its children.

account_analytic_parent/wizard/account_analytic_wizard.py
j-zaballa: Hierarchical tree view is no more: [GitHub thread](https://github.com/odoo/odoo/issues/20659).

I can not think of a way to replicate the hierarchical view. So, maybe just get rid of the wizard, the report view and action and the test?

account_analytic_parent/tests/test_account_analytic_account.py
```diff
@@ -0,0 +1,82 @@
...
+        self.assertEqual(self.analytic_son.debit, 0,
+                         "Analytic account in the debit side")
+        self.assertEqual(self.analytic_son.credit, 50, "Wrong amount")
+        self.assertEqual(self.analytic_son.balance, 50, "Wrong amount")
+        self.assertEqual(self.analytic_parent2.debit, 0,
+                         "Analytic account in the debit side")
+        self.assertEqual(self.analytic_parent2.credit, 50, "Wrong amount")
+        self.assertEqual(self.analytic_parent2.balance, 50, "Wrong amount")
+
+    def test_wizard(self):
```
j-zaballa: In case the wizard and the account chart view are deleted, delete this too.

account_analytic_parent/views/account_analytic_account_view.xml
```diff
@@ -0,0 +1,52 @@
...
+        <field name="type">form</field>
+        <field name="inherit_id" ref="analytic.view_account_analytic_account_form"/>
+        <field name="arch" type="xml">
+            <field name="tag_ids" position="after">
+                <field name="parent_id" string="Parent Analytic Account"/>
+            </field>
+        </field>
+    </record>
+
+    <record id="view_account_analytic_account_report" model="ir.ui.view">
```
j-zaballa: As `field_parent` is no longer supported, get rid of this view, unless someone can think of a way to replicate this in another way.

account_analytic_parent/views/account_analytic_account_view.xml
```diff
@@ -0,0 +1,52 @@
...
+                <field name="company_id" groups="base.group_multi_company"/>
+                <field name="debit"/>
+                <field name="credit"/>
+                <field name="balance"/>
+            </tree>
+        </field>
+    </record>
+
+    <record id="action_analytic_account_report"
+            model="ir.actions.act_window">
```
j-zaballa: Delete this action if the account chart report view is deleted.

j-zaballa (commented):
All seems ok apart from the fact that `field_parent` is not supported anymore :)

account_analytic_parent/wizard/account_analytic_wizard.py
jaumeplanas: Since hierarchical tree view has gone, I think also it would be better to discard this functionality and not to be maintained anymore.
If the community agree, I'll discard this useless part of the code in the next days, so this PR could be approved and merged.

account_analytic_parent/models/account_analytic_account.py
jaumeplanas: I think the code is ok as it is too.

j-zaballa (approved):
This should be merged IMHO

PR #171, opened 2018-05-15
[FIX] product_analytic: Pass integration tests

yajo:
This addon fails if `account_analytic_required` addon is preinstalled,
with:

    Traceback (most recent call last):
    `   File "/opt/odoo/auto/addons/product_analytic/tests/test_account_invoice.py", line 32, in setUp
    `     'type': 'other'})
    `   File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3847, in create
    `     record = self.browse(self._create(old_vals))
    `   File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3942, in _create
    `     cr.execute(query, tuple(u[2] for u in updates if len(u) > 2))
    `   File "/opt/odoo/custom/src/odoo/odoo/sql_db.py", line 154, in wrapper
    `     return f(self, *args, **kwargs)
    `   File "/opt/odoo/custom/src/odoo/odoo/sql_db.py", line 231, in execute
    `     res = self._obj.execute(query, params)
    ` IntegrityError: null value in column "analytic_policy" violates not-null constraint
    ` DETAIL:  Failing row contains (26, 1, Test account type, 1, null, 2018-05-15 08:07:36.153701, 2018-05-15 08:07:36.153701, f, other, null).

@Tecnativa

### DIFF
--- a/account_analytic_default_account/models/account_analytic_default_account.py
+++ b/account_analytic_default_account/models/account_analytic_default_account.py
@@ -2,6 +2,7 @@
 # Copyright 2018 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
+from odoo.tools import config
 from odoo import api, fields, models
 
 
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Travis is red

yajo:
Please hold on the merge until I confirm it works in the integration environment. :hand:

yajo:
Indeed this works in integration environment. I forgot to mention.

chienandalu (approved):

yajo:
Ready to merge

yajo:
Travis' :x: comes from the fact that [`product_analytic` changes `_onchange_product_id()`](https://github.com/OCA/account-analytic/blob/2d5c1ebd469f8795b3c56dadd9286b382b9f4cfa/product_analytic/models/account_invoice.py#L20-L29) in a way that is incompatible with [the way it's modified in `account_analytic_default_account`](https://github.com/OCA/account-analytic/blob/2d5c1ebd469f8795b3c56dadd9286b382b9f4cfa/account_analytic_default_account/models/account_analytic_default_account.py#L80-L90).

So how do I fix this?

rafaelbn (approved):
Then :+1:  please @pedrobaeza could you answer @Yajo ?

yajo:
I just want to point out that the current implementation breaks the integration environment we have, which was after all the main purpose of this PR. I'd go for the test skipping solution.

pedrobaeza:
Let me check first to solve the tests itself.

yajo:
Sorry forget about https://github.com/OCA/account-analytic/pull/171#issuecomment-396961953

This implementation works, although we don't use all addons from this branch in that environment.

pedrobaeza:
Now this is properly fixed. As always, there are 2 overlapping behaviors (on the onchange of the product in the invoice line) that chokes when making tests. Adding the code for inhibiting both codes when we are not testing them leads to a green result.

There were no problem before as one of the conflicting modules was not loaded before testing the other.

PR #147, opened 2018-01-03
[IMP] account_analytic_distribution: Add analytic tags in distribution

carlosdauden:
@Tecnativa

PR #179, opened 2018-06-12
[10.0][IMP] purchase_procurement_analytic: Take into account bare pro…

rousseldenis:
…curement

If purchase orders were generated through bare procurement orders, generated moves do not
contain the analytic account set on procurement.

### DIFF
--- a/purchase_procurement_analytic/__manifest__.py
+++ b/purchase_procurement_analytic/__manifest__.py
@@ -7,7 +7,7 @@
     'name': 'Purchase Procurement Analytic',
     'summary': 'This module sets analytic account in purchase order line from '
                'procurement analytic account',
-    'version': '10.0.1.0.2',
+    'version': '10.0.2.0.0',
     'category': 'Analytic',
     'license': 'AGPL-3',
     'author': "Tecnativa, "
@@ -16,6 +16,7 @@
     'depends': [
         'purchase',
         'procurement_analytic',
+        'stock_analytic',
     ],
     'installable': True,
 }
--- a/purchase_procurement_analytic/README.rst
+++ b/purchase_procurement_analytic/README.rst
@@ -53,6 +53,7 @@ Contributors
 * Carlos Dauden <carlos.dauden@tecnativa.com>
 * Pedro M. Baeza <pedro.baeza@tecnativa.com>
 * Vicent Cubells <vicent.cubells@tecnativa.com>
+* Denis Roussel <denis.roussel@acsone.eu>
 
 Maintainer
 ----------
--- a/purchase_procurement_analytic/models/__init__.py
+++ b/purchase_procurement_analytic/models/__init__.py
@@ -3,4 +3,5 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import procurement
+from . import purchase_order_line
 from . import stock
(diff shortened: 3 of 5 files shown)


rousseldenis:
@Tecnativa @aheficent

rafaelbn:
@carlosdauden @pedrobaeza @cubells could you review here as previous contributors please?

purchase_procurement_analytic/__manifest__.py
```diff
@@ -16,6 +16,7 @@
     'depends': [
         'purchase',
         'procurement_analytic',
+        'stock_analytic',
```
pedrobaeza, line 19: I don't understand why you need this new dependency

AaronHForgeFlow:
The stock_account is passing the analytic account from the procurement order to the stock move. In this case the one that passes the analytic account to the stock move is the purchase order, isn't it?

rousseldenis:
@aheficent @pedrobaeza Yes that's the point.
Not to duplicate the code from stock_analytic here, we have to depends on it.

This PR's intent is to fill a gap in this module to be complete.

The existing module takes into account moves that generates procurements. This works only if another procurement or ... had generated a stock move.

If you make a purchase order manually, as it does not generate procurement, the analytic account is not present in stock moves

purchase_procurement_analytic/__manifest__.py
cubells: @rousseldenis me too.
I do tests without stock_analytic dependency and module works fine. Why this dependency?

AaronHForgeFlow (commented):
For me this is fine. Maybe we can move it to a third module that depends on stock_analytic and purchase_procurement_analytic...

AaronHForgeFlow (approved):
:+1: Still same opinion

AaronHForgeFlow:
@rousseldenis can you rebase? last commit is quite old.

AaronHForgeFlow:
@cubells are you ok with this: https://github.com/OCA/account-analytic/pull/179#issuecomment-396870525 ?

rousseldenis:
@aheficent Rebase done

AaronHForgeFlow:
@rousseldenis Thank you.

@pedrobaeza as `stock_analytic` module adds `analytic_account_id` field in `stock.move` and here we are propagating the analytic account to the stock move, it makes sense to add the dependency here.

rousseldenis:
https://github.com/OCA/account-analytic/pull/147 merging caused Travis error.

rousseldenis:
This one is ready

purchase_procurement_analytic/__manifest__.py
```diff
@@ -7,7 +7,7 @@
     'name': 'Purchase Procurement Analytic',
     'summary': 'This module sets analytic account in purchase order line from '
                'procurement analytic account',
-    'version': '10.0.1.0.2',
+    'version': '10.0.1.0.3',
```
MiquelRForgeFlow: -> 10.0.2.0.0, because you are adding a new dependency

rousseldenis:
@mreficent Done

MiquelRForgeFlow (approved):

AaronHForgeFlow:
@pedrobaeza  do you consider your comments are attended here?

pedrobaeza:
OK, although it's not ideal to change dependencies in a stable version.

PR #183, opened 2018-07-10
[FIX] *procurement_analytic: duplicated account analytic field in view

cubells:
Twice was added account_analytic_field to the same view.
Update view to merge both changes.
Update README in other modules that depends on procurement analytic.

cc @Tecnativa


pedrobaeza:
You are not changing any view in  sale_procurement_analytic

cubells:
No, I'm not. Views are in procurement_analytic.

pedrobaeza:
Then you shouldn't change the version of that modules. Changing a bit the README doesn't require to change module version number. Any way, I'm wondering why do you remove the "Other information" tab sentence. Are you sure this is not needed? Maybe you have an extra module that shows that information.

cubells:
Because field analytic_account_id is duplicated in form view. It's shown in two locations, in form header and in other information tab. So, with this changes one location disappear. I choose that header location remains in view and remove from other information tab.

cubells:
So README information must be updated in other modules.

pedrobaeza:
This seems a bit strange to not being detected before, but OK if you say so. What I'm also saying is that you don't need to change module version number for only changing the README.

cubells:
@pedrobaeza undo change of versions.

pedrobaeza (approved):

AaronHForgeFlow (approved):
:+1: I agree with this

rousseldenis (approved):

PR #190, opened 2018-10-01
[11] [MIG] Account Analytic No Lines

ThomasBinsfeld:
Migration of account_analytic_no_lines from 10.0 to 11.0.
#133 

Migration to 12.0 : #191

PR #191, opened 2018-10-01
[12] [MIG] Account Analytic No Lines

ThomasBinsfeld:
Migration of account_analytic_no_lines from 11.0 to 12.0.
11.0 migration : #190 
#189


account_analytic_no_lines/models/account_analytic_account.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class AccountAnalyticAccount(models.Model):
+
+    _inherit = 'account.analytic.account'
+
+    @api.multi
+    def _compute_gl_debit_credit_balance(self):
+        aml_obj = self.env['account.move.line']
+        domain = [('analytic_account_id', 'in', self.ids)]
```
espo-tony, line 14: For sake of inheritance simplicity it could be better to put the domain creation in a separate method and invoke it here.

account_analytic_no_lines/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+
+{
+    'name': 'Account Analytic No Lines',
+    'summary': """
+        Hide analytics lines and disable
+        their generation from a move line.""",
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'https://acsone.eu/',
```
nikul-serpentcs: @ThomasBinsfeld Replace URL as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10)

account_analytic_no_lines/views/account_move_line.xml
```diff
@@ -0,0 +1,35 @@
...
+-->
+
+<odoo>
+
+<record id="view_move_line_form" model="ir.ui.view">
+    <field name="name">hide_analytic_lines.account.move.line.form</field>
+    <field name="model">account.move.line</field>
+    <field name="inherit_id" ref="account.view_move_line_form"/>
+    <field name="arch" type="xml">
+        <xpath expr="//field[@name='analytic_line_ids']/.." position="attributes">
```
nikul-serpentcs, line 14: Directly use `field`  here

nikul-serpentcs (commented):
Please squash commits, Improve Code

account_analytic_no_lines/views/account_move_line.xml
ThomasBinsfeld: This xpath doesn't target the field but the group just before (which has no name).

account_analytic_no_lines/models/account_analytic_account.py
ThomasBinsfeld: @espo-tony I won't refactor the code in this migration PR. We can do that in a refactoring PR after the merge.

account_analytic_no_lines/models/account_move_line.py
```diff
@@ -0,0 +1,24 @@
...
+            return False
+        if not hasattr(move_line.product_id, 'expense_policy'):
+            return False
+        if move_line.product_id.expense_policy == 'no':
+            return False
+        return True
+
+    @api.multi
+    def create_analytic_lines(self):
+        self = self.filtered(self._must_create_analytic_line)
```
florian-dacosta, line 23: @ThomasBinsfeld 
The method is never actually called here right?
it should be `self = self.filtered(self._must_create_analytic_line())`
And `_must_create_analytic_line` should have an api.multi decorator and manage multiples account move line. (and remove the `move_line` parameter)

That said, I wonder, why not make it simpler and just totally by pass `create_analytic_lines`?
I am not really confortable with the "expense policy" of the products, but I wonder, it would not make sense for someone who need analytic lines for products with special expense policies to install this module right?

Problem is also present in version 10 and 11.

florian-dacosta (changes requested):

PR #193, opened 2018-10-10
[10.0][IMP]account_analytic_parent. Propagate active through hierarchy

AaronHForgeFlow:
If the parent analytic account is archived then all the children accounts should be archived. The same the other way around. 

When manual archiving the accounts it is common to miss some of the children causing confusion to see archived accounts appearing again.

Perhaps the parents should be also archived, but I am not sure of that...

### DIFF
--- a/account_analytic_parent/models/account_analytic_account.py
+++ b/account_analytic_parent/models/account_analytic_account.py
@@ -67,8 +67,8 @@ class AccountAnalyticAccount(models.Model):
             if (account.active and account.parent_id and not
                     account.parent_id.active):
                 raise UserError(
-                    _('Please activate first parent account')
-                    % self.parent_id.display_name)
+                    _('Please activate first parent account %s')
+                    % account.parent_id.display_name)
 
     @api.multi
     def archive_account(self):
--- a/account_analytic_parent/models/account_analytic_account.py
+++ b/account_analytic_parent/models/account_analytic_account.py
@@ -61,6 +61,16 @@ class AccountAnalyticAccount(models.Model):
         return res
 
     @api.multi
+    @api.constrains('active')
+    def check_parent_active(self):
+        for account in self:
+            if (account.active and account.parent_id and not
+                    account.parent_id.active):
+                raise UserError(
+                    _('Please activate first parent account')
+                    % self.parent_id.display_name)
+
+    @api.multi
     def archive_account(self):
         for account in self:
             account.active = False
--- a/account_analytic_parent/models/account_analytic_account.py
+++ b/account_analytic_parent/models/account_analytic_account.py
@@ -59,3 +59,14 @@ class AccountAnalyticAccount(models.Model):
                 current = current.parent_id
             res.append((account.id, name))
         return res
+
+    @api.multi
+    def archive_account(self):
+        for account in self:
+            account.active = False
+
+    @api.multi
+    def write(self, vals):
+        if 'active' in vals and not vals['active']:
+            self.mapped('child_ids').archive_account()
+        return super(AccountAnalyticAccount, self).write(vals)
--- a/account_analytic_parent/tests/test_account_analytic_account.py
+++ b/account_analytic_parent/tests/test_account_analytic_account.py
@@ -81,3 +81,9 @@ class TestAccountAnalyticRecursion(TransactionCase):
         result = self.wizard.analytic_account_chart_open_window()
         self.assertTrue('2017-01-01' in result['context'])
         self.assertTrue('2017-12-31' in result['context'])
+
+    def test_archive(self):
+        self.analytic_parent1.toggle_active()
+        self.assertEqual(self.analytic_son.active, False)
+        self.analytic_parent1.toggle_active()
+        self.assertEqual(self.analytic_son.active, False)
--- a/account_analytic_parent/tests/test_account_analytic_account.py
+++ b/account_analytic_parent/tests/test_account_analytic_account.py
@@ -87,3 +87,6 @@ class TestAccountAnalyticRecursion(TransactionCase):
         self.assertEqual(self.analytic_son.active, False)
         self.analytic_parent1.toggle_active()
         self.assertEqual(self.analytic_son.active, False)
+        self.analytic_parent1.toggle_active()
+        with self.assertRaises(ValidationError):
+            self.analytic_son.toggle_active()


pedrobaeza:
I'm not sure about the contrary. If you archive some children accounts before archiving the parent, you shouldn't activate them when activating the parent.

AaronHForgeFlow:
@pedrobaeza  hmm yes, maybe you want to unarchive some child for specific reason. That is fine to me and it is more generic.

In the other hand, the action should be easy to be reverted, so the opposite action just undo what the first did... still unsure.

mreficent:
To me it doesn't make sense to have a child active and its parent inactive.

account_analytic_parent/models/account_analytic_account.py
```diff
@@ -59,3 +59,9 @@ def name_get(self):
                 current = current.parent_id
             res.append((account.id, name))
         return res
+
+    @api.multi
+    def write(self, vals):
+        if 'active' in vals:
+            self.mapped('child_ids').toggle_active()
+        return super(AccountAnalyticAccount, self).write(vals)
```
MiquelRForgeFlow, line 82: I think it should be this way:
```python
    @api.multi
    def write(self, vals):
        if 'active' in vals and not vals['active']:
            self.mapped('child_ids').write({'active': False})
        return super(AccountAnalyticAccount, self).write(vals)
```
because activating the parent doesn't mean you want activate all its children

MiquelRForgeFlow:
To me it doesn't make sense to have a child active and its parent inactive, so I think you should guarantee that by adding:
```python
     @api.multi
     @api.constrains('active')
     def check_parent_active(self):
         for account in self:
             if account.active and account.parent_id and not account.parent_id.active:
                 raise UserError(
                     _('You can not activate this analytic account.\n'
                       'You should activate first its parent account %s.') % self.parent_id.display_name
                 )
    )
```

account_analytic_parent/models/account_analytic_account.py
MiquelRForgeFlow: edited

AaronHForgeFlow:
@mreficent  That is why maybe it's  better to activate the children if the parent is activated.  Imagine if we archive by mistake a big analytic account with many children and then we have unarchive all of them. It is a  pain.

account_analytic_parent/models/account_analytic_account.py
AaronHForgeFlow: It has to be with toggle_active because calling write will get recursion error.

MiquelRForgeFlow: I don't see the recursion error. Please put me the recursion error

MiquelRForgeFlow:
Thus avoid archiving my mistake. When archiving, you could call a mini wizard to ask the user if they are sure about archiving a big account with XXX children and adverting the user that archiving the account will archive all its children. Another way is restricting the people able to archive analytic accounts.

MiquelRForgeFlow:
Another way is when activating an account, a mini wizard is shown asking if the user wants to activate just only the account or also the account and all its children and let the user choose.

account_analytic_parent/models/account_analytic_account.py
AaronHForgeFlow: I am not just saying, I tested that.

AaronHForgeFlow:
I am not entirely convinced but I'll go with the majority. No unarchiving the children with the last commit. I prefer to leave it like this, no warnings or similar.

account_analytic_parent/models/account_analytic_account.py
MiquelRForgeFlow: ok, then do instead:

```python

    @api.multi
    def archive_account(self):
        for account in self:
            record.active = False

    @api.multi
    def write(self, vals):
        if 'active' in vals and not vals['active']:
            self.mapped('child_ids').archive_account()
        return super(AccountAnalyticAccount, self).write(vals)
```

to avoid toggle_active, because if you want to archive an account, you wouldn't want to activate its already archived children

AaronHForgeFlow: yes, I forgot to change also that.  Thanks.

MiquelRForgeFlow:
please, add the constrain method :)

account_analytic_parent/models/account_analytic_account.py
```diff
@@ -59,3 +59,14 @@ def name_get(self):
...
+    @api.multi
+    def archive_account(self):
+        for account in self:
+            account.active = False
+
+    @api.multi
+    def write(self, vals):
+        if 'active' in vals and not vals['active']:
+            self.mapped('child_ids').archive_account()
+        return super(AccountAnalyticAccount, self).write(vals)
```
MiquelRForgeFlow, line 82: perhaps you prefer this

```python
@api.multi
    def archive_account(self, unarchive=False):
        for account in self:
            account.active = unarchive

    @api.multi
    def write(self, vals):
        if 'active' in vals and not vals['active']:
            self.mapped('child_ids').archive_account()
        res = super(AccountAnalyticAccount, self).write(vals)
        if 'active' in vals and vals['active'] and self.env.context('unarchive_children', False):
            self.mapped('child_ids').archive_account(True)
        return res
```

This way the unarchiving of all children is also possible, but you shall pass a context to do that 😏😉
(the unarchiving of children is done after the unarchive of the parent to make this compatible with the constrain method I put you before)

AaronHForgeFlow: It's ok, better to leave it like this. I can overwrite this in custom module.

MiquelRForgeFlow (approved):

AaronHForgeFlow:
Closing. Wrong approach here, causing more confusion than helping.

MiquelRForgeFlow:
I don't really understand why closing. If not doing this kind of fix or someone similar, people could potentially archive and unarchive accounts without noticing if their parent and child accounts are archived or not.

AaronHForgeFlow:
If archiving the parent and not the children you can still archive the children not an issue. As you can archive / Unarchive in batch users can archive several projects at the same time. If I add the constraint and I try to archive a parent and the children in batch I probably get the constraint raised because Odoo is archiving a child first and not the parent. So, it is screwing the batch prosessing feature.

MiquelRForgeFlow:
No, in that case is you mention is not raised because that write is properly built to avoid the constraint. You can test it.

AaronHForgeFlow:
I tested and I got what I described in my previous message.

PR #196, opened 2018-10-19
[10.0][IMP] account_analytic_parent: propagate active through hierarchy

MiquelRForgeFlow:
Fix of https://github.com/OCA/account-analytic/pull/193.

@aheficent could you test it, please?

### DIFF
--- a/account_analytic_parent/models/account_analytic_account.py
+++ b/account_analytic_parent/models/account_analytic_account.py
@@ -74,3 +74,20 @@ class AccountAnalyticAccount(models.Model):
                 current = current.parent_id
             res.append((account.id, name))
         return res
+
+    @api.multi
+    @api.constrains('active')
+    def check_parent_active(self):
+        for account in self:
+            if (account.active and account.parent_id and
+                    account.parent_id not in self and
+                    not account.parent_id.active):
+                raise UserError(
+                    _('Please activate first parent account %s')
+                    % account.parent_id.display_name)
+
+    @api.multi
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):
:+1: Functional tests. Thanks.

LoisRForgeFlow (approved):
LGTM :+1:

MiquelRForgeFlow:
@pedrobaeza

account_analytic_parent/models/account_analytic_account.py
```diff
@@ -59,3 +59,25 @@ def name_get(self):
...
+
+    @api.multi
+    def archive_account(self):
+        for account in self:
+            account.active = False
+
+    @api.multi
+    def write(self, vals):
+        if 'active' in vals and not vals['active']:
+            self.mapped('child_ids').archive_account()
```
pedrobaeza:
```suggestion
            self.mapped('child_ids').write({'active': False})
```
and remove extra method

pedrobaeza (changes requested):
This doesn't propagate active to grandson and so

MiquelRForgeFlow:
It propagates, because when you do `self.mapped('child_ids').write()`, then it will enter again in the `write` but for their children and thus will execute a `self.mapped('child_ids').write()` for the grandson and so on.

pedrobaeza:
Right, thanks for the note.

MiquelRForgeFlow:
ready :green_apple:

PR #197, opened 2018-10-23
[10.0][FIX] product_analytic_purchase: Only is executed if exist product

igallart:
If product is empty, _get_product_analytic_accounts from product_analytic module fails

### DIFF
--- a/product_analytic_purchase/README.rst
+++ b/product_analytic_purchase/README.rst
@@ -41,6 +41,7 @@ Contributors
 ------------
 
 * Alexis de Lattre <alexis.delattre@akretion.com>
+* Isaac Gallart <igallart@puntsistemes.es>
 
 Maintainer
 ----------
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I don't get why the previous condition is not ok. You are basically making the same check.

igallart:
> I don't get why the previous condition is not ok. You are basically making the same check.

@pedrobaeza, not is the same, when is object new_id created from onchange from another module, it can continue, but is a fake object, and your template is empty and it fails.

pedrobaeza:
No, that can't be, unless you create that product incorrectly in the onchange in a custom module.

igallart:
> No, that can't be, unless you create that product incorrectly in the onchange in a custom module.

It fail when this module is executed first.
https://github.com/OCA/product-variant/blob/10.0/purchase_variant_configurator_on_confirm/models/purchase_order.py#L46

pedrobaeza:
Then you should fix that one instead.

igallart:
Thanks @pedrobaeza, I don't analyze correctly the modules. I will fix in purchase_variant_configurator_on_confirm

Issue #198, opened 2018-10-26
[10.0]account_analytic_parent. Debit, credit and balance computation is super slow.


AaronHForgeFlow:
When overriding the method that computes those fields here: https://github.com/OCA/account-analytic/blob/10.0/account_analytic_parent/models/account_analytic_account.py#L24 Odoo calls the same method for all the children and it gets extremely slow when the number of analytic accounts grow, specially with many children.

We can make those fields stored but then we have to make sure it is recomputed everytime its needed and it's complex.

Any ideas are welcome! :hugs:

pedrobaeza:
Use `read_group` for computing them. Your colleague @mreficent made something similar in contracts.

AaronHForgeFlow:
Ok thanks, @mreficent can you point me to that change? I could not find it in contracts.

MiquelRForgeFlow:
https://github.com/OCA/contract/pull/170/commits/402370f1cbe5cde6b484200f87596e9d803a36af#diff-3dc0d52ea9449a93459ef5b7fc36a060R19

AaronHForgeFlow:
Thanks @mreficent

AaronHForgeFlow:
Cannot group_by non stored fields. And read group does not return the values that are not in the group_by. I'll go for a query instead.

edit:I can read_group the analytic lines but that would be duplicating the core method

pedrobaeza:
@aheficent this has nothing to do with stored/non stored. Please use `read_group` properly.

AaronHForgeFlow:
Do you mean to read_group in the analytic lines and not call super instead?
update: I think I got it...

pedrobaeza:
Well, you limit the read_group in the domain to the `self.ids` scope.

PR #200, opened 2018-10-29
[10.0][IMP]performance account_analytic_parent

AaronHForgeFlow:
solves issue #198 

Thanks @pedrobaeza  and @mreficent for the help!

### DIFF
--- a/account_analytic_parent/models/account_analytic_account.py
+++ b/account_analytic_parent/models/account_analytic_account.py
@@ -32,7 +32,7 @@ class AccountAnalyticAccount(models.Model):
         for account in self.filtered(lambda x: x.child_ids):
             domain = [('account_id', 'child_of', account.ids)]
             credit_groups = analytic_line_obj.read_group(
-                domain=domain + [('amount', '>=', 0.0)],
+                domain=domain + [('amount', '>', 0.0)],
                 fields=['account_id', 'amount'],
                 groupby=['account_id']
             )
--- a/account_analytic_parent/models/account_analytic_account.py
+++ b/account_analytic_parent/models/account_analytic_account.py
@@ -27,10 +27,25 @@ class AccountAnalyticAccount(models.Model):
         of analytic account changes
         """
         super(AccountAnalyticAccount, self)._compute_debit_credit_balance()
-        for account in self:
-            account.debit += sum(account.mapped('child_ids.debit'))
-            account.credit += sum(account.mapped('child_ids.credit'))
-            account.balance += sum(account.mapped('child_ids.balance'))
+        analytic_line_obj = self.env['account.analytic.line']
+        # compute only analytic line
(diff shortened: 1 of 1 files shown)


mreficent:
:red_circle:

MiquelRForgeFlow (approved):
LGTM

account_analytic_parent/models/account_analytic_account.py
```diff
@@ -27,10 +27,25 @@ def _compute_debit_credit_balance(self):
...
         """
         super(AccountAnalyticAccount, self)._compute_debit_credit_balance()
-        for account in self:
-            account.debit += sum(account.mapped('child_ids.debit'))
-            account.credit += sum(account.mapped('child_ids.credit'))
-            account.balance += sum(account.mapped('child_ids.balance'))
+        analytic_line_obj = self.env['account.analytic.line']
+        # compute only analytic line
+        for account in self.filtered(lambda x: x.child_ids):
+            domain = [('account_id', 'in', account.child_ids.ids)]
```
pedrobaeza: operator should be `child_of` for including account itself.

pedrobaeza: and do the test over `self.ids`, not `self.child_ids.ids`

AaronHForgeFlow: account itself is not included in the loop but it was already included when calling super, then in the loop we increment the values by adding the children ones. But maybe is less confusing doing that way, it's ok.

AaronHForgeFlow: This is resolved, isn't it?

AaronHForgeFlow: @pedrobaeza Can you please update your review? IMO your comment is attended.

account_analytic_parent/models/account_analytic_account.py
```diff
@@ -27,10 +27,25 @@ def _compute_debit_credit_balance(self):
...
-        for account in self:
-            account.debit += sum(account.mapped('child_ids.debit'))
-            account.credit += sum(account.mapped('child_ids.credit'))
-            account.balance += sum(account.mapped('child_ids.balance'))
+        analytic_line_obj = self.env['account.analytic.line']
+        # compute only analytic line
+        for account in self.filtered(lambda x: x.child_ids):
+            domain = [('account_id', 'child_of', account.ids)]
+            credit_groups = analytic_line_obj.read_group(
+                domain=domain + [('amount', '>=', 0.0)],
```
pedrobaeza: No need to include 0 results

pedrobaeza: I mean they are not going to add anything to the sum, as their amount is 0, so better to strip them.

AaronHForgeFlow: yes, it makes sense

PR #204, opened 2018-11-14
[11.0] [MIG] sale/purchase_procurement_analytic > procurement_mto_analytic

gittusr:


aisopuro:
Hi @gittusr . I'm not extremely familiar with this module construction, but based on the READMEs in the 10.0 branch, it seems like the original way these modules worked was that you had a "core module" (`procurement_analytic`) which allowed you to have an analytic account as part of the procurement chain. Other modules could then hook into this logic as they wanted: `sale_procurement_analytic` could add the sale's analytic account to the procurement, and `purchase_procurement_analytic` could read it if it was present.

The procurement logic has changed a lot from 10.0 to 11.0, and especially the `procurement.order` model has disappeared, so there isn't really a specific model designed to handle procurements anymore. 

I think it would be good if you could explicitly mention this logic change and the reasons why you removed all the other modules in the commit message so that there is an explanation in the history of why the modules disappeared.

I also don't know what the policy is, but since this "new" module combines `sale` and `purchase` functionality, I feel like it should perhaps have a new name. `sale_purchase_procurement_analytic` perhaps?

aisopuro (approved):
Look good to me. :ok_hand:

AaronHForgeFlow (approved):
Functional test :+1: 

As there is an object called procurement rule it is ok to me to keep the module name.

pedrobaeza (changes requested):
I agree with @aisopuro that the scope has changed and thus, you have to rename the module name. Basically, as I understand, this one gets rid out the previous 3 modules: `procurement_analytic`, `sale_procurement_analytic` and `purchase_procurement_analytic`. 

So for me the most reasonable option is to let the name of this in just `procurement_analytic`, or go a bit further with `procurement_mto_analytic`, because at the end, this is used for MTO only.

pedrobaeza (approved):
Squashing migration commits (and maybe you can generate the README already following https://github.com/OCA/maintainer-tools/#readme-generator

gurneyalex:
I'm closing and reopening to test the clabot.

okuryan:
@pedrobaeza , we see that Pull request is marked as approved, but still see that there are some issues with it. Can @gittusr  help somehow? Any actions required from us?

pedrobaeza:
I have rebased the branch and did following things:

* Restore authorship and contributors. @gittusr you can't remove our authorship from the module. I have respected yours as you have touched enough the module for granting it though.
* Generated README.rst from fragments.
* Correct URL in manifest.
* Squashed commits.

If everything goes right, I'll merge.

PR #209, opened 2018-12-04
[12.0] [MIG] analytic_base_department

anandkansagra:
Issue [Reference](https://github.com/OCA/account-analytic/issues/189)


emagdalenaC2i (approved):
Code review

analytic_base_department/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2011-2018 Camptocamp SA
+# Copyright 2013 Daniel Reis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Base Analytic Department Categorization",
+    "version": "12.0.1.0.0",
+    "author": "Camptocamp, Daniel Reis, Odoo Community Association (OCA)",
```
nikul-serpentcs, line 7: @anandkansagra `Daniel Reis`  name in an author.
@dreispt can you please send company name?

dreispt: There is no company name to go there.

analytic_base_department/__manifest__.py
```diff
@@ -3,13 +3,12 @@
...
 {
     "name": "Base Analytic Department Categorization",
-    "version": "11.0.1.0.0",
+    "version": "12.0.1.0.0",
     "author": "Camptocamp, Daniel Reis, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Generic Modules/Projects & Services",
     "website": "https://github.com/OCA/account-analytic",
     "depends": ["account",
-                "analytic",
```
dreispt: Why is this removed? The module uses `account.analytic.account` Model, provided by the `analytic` module.

analytic_base_department/views/analytic.xml
```diff
@@ -1,70 +1,70 @@
...
+    <record id="view_account_analytic_account_form" model="ir.ui.view">
+        <field name="name">account.analytic.account.form</field>
+        <field name="model">account.analytic.account</field>
+        <field name="inherit_id" ref="analytic.view_account_analytic_account_form"/>
+        <field name="arch" type="xml">
+            <field name="code" position="after">
+                <field name="department_id"/>
+            </field>
+        </field>
+    </record>
```
dreispt, line 13: No point in changing the XML indentation, but discussing that would be bikeshedding.

analytic_base_department/__manifest__.py
anandkansagra: @dreispt ,

If we will install the `account` module then `analytic` module will automatically installed due to dependency. So no need to put `analytic` module in the dependency
Here is the Reference:- https://github.com/odoo/odoo/blob/12.0/addons/account/__manifest__.py#L18

dreispt (commented):
Thanks @anandkansagra, good catch. I would then prefer to keep the closest dependency: this needs `analytic` but doesn't need `account`. So I would rather remove the `account` dependency.

nikul-serpentcs (approved):
Code Review LGTM :+1:

anandkansagra:
> Thanks @anandkansagra, good catch. I would then prefer to keep the closest dependency: this needs `analytic` but doesn't need `account`. So I would rather remove the `account` dependency.

@dreispt ,
Technically it will not throw an error. I have seen that in the `analytic` module, there seems no menuitem. All the menuitem related `analytic` module is defined in the `account` module. So do we need to change it in this PR?

pedrobaeza:
I have squashed a bit commit history and merging when finished.

PR #215, opened 2019-01-03
[IMP] account_analytic_parent: improve account tree view

cubells:
Now, account analytic tree view is most readable:

parent
cc @Tecnativa

### DIFF
--- a/account_analytic_parent/models/account_analytic_account.py
+++ b/account_analytic_parent/models/account_analytic_account.py
@@ -11,10 +11,25 @@ from odoo.exceptions import UserError
 
 class AccountAnalyticAccount(models.Model):
     _inherit = "account.analytic.account"
+    _parent_name = "parent_id"
+    _parent_store = True
+    _parent_order = 'name'
+    _rec_name = 'display_name'
+    _order = 'parent_left'
 
+    parent_left = fields.Integer(
+        string='Left Parent',
+        index=True,
+    )
+    parent_right = fields.Integer(
+        string='Right Parent',
+        index=True,
+    )
     parent_id = fields.Many2one(
         comodel_name="account.analytic.account",
-        string="Parent Analytic Account"
+        string="Parent Analytic Account",
+        index=True,
+        ondelete='cascade',
     )
     child_ids = fields.One2many(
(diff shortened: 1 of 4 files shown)


account_analytic_parent/models/account_analytic_account.py
```diff
@@ -11,16 +11,38 @@
...
     _inherit = "account.analytic.account"
+    _parent_name = "parent_id"
+    _parent_store = True
+    _parent_order = 'name'
+    _rec_name = 'complete_name'
+    _order = 'parent_left'
+
+    parent_left = fields.Integer('Left Parent', index=1)
+    parent_right = fields.Integer('Right Parent', index=1)
+    complete_name = fields.Char(
```
pedrobaeza: Why adding this field if we have `display_name` for the same purpose?

cubells: You are right @pedrobaeza . I've just removed field and method.

account_analytic_parent/models/account_analytic_account.py
```diff
@@ -11,10 +11,25 @@
...
     _inherit = "account.analytic.account"
+    _parent_name = "parent_id"
+    _parent_store = True
+    _parent_order = 'name'
+    _rec_name = 'display_name'
+    _order = 'parent_left'
 
+    parent_left = fields.Integer(
+        string='Left Parent',
+        index=1,
```
pedrobaeza: Why `index=1` instead of  `index=True`? It shouldn't be a problem, but better to be homogeneous

pedrobaeza (approved):

cubells:
Balance and debit values are incorrect in parents.
In order to compute correct values in accounts and child I added a change.
@pedrobaeza can you check again?

pedrobaeza:
Yeah, good catch! I'm going to fast-track this due to this bug.

pedrobaeza:
@alexey-pelykh you would be interested in this for v12, although the implementation of hierarchy is a slight different.

alexey-pelykh:
@pedrobaeza thanks, addressed in #225

PR #216, opened 2019-01-10
[FIX] procurement_mto_analytic: added missing dependency

aisopuro:
Tests earlier likely passed because the dependency is in the
same repository, and so the CI would install both modules for running
tests, thus making it seem like it should work.


aisopuro:
Misunderstanding/uncertainty: it seems the `stock_analytic` module is not necessary after all, nor does it do what I assumed it does.

PR #236, opened 2019-05-21
[11.0][FIX] account_analytic_distribution travis warn

tbaden:
duplicate view definition in the files:
account_move_view.xml
account_move_line_view.xml

removed it from account_move_line_view.xml as the model used for the view is account.move
and account.move.line has already the view view_move_line_form for the field analytic_distribution_id


pedrobaeza (approved):
Fast-tracking this

PR #237, opened 2019-05-22
[FIX] modify domain search to provide consistency with account

i-vyshnevska:
In case of reconciliation default account is not searched correctly what leads to errors during it
steps to reproduce it:

1. add policy 'posted' in journal of currency differences 
2. setup default aal to accounts + define company to apply
3. make payment in different currencies to trigger currency difference computation

### DIFF
--- a/account_analytic_default_account/models/account_analytic_default_account.py
+++ b/account_analytic_default_account/models/account_analytic_default_account.py
@@ -77,27 +77,36 @@ class AccountAnalyticDefaultAccount(models.Model):
 class AccountInvoiceLine(models.Model):
     _inherit = "account.invoice.line"
 
-    @api.onchange('product_id', 'account_id')
+    @api.onchange('product_id', 'account_id', 'partner_id', 'company_id',
+                  'date')
     def _onchange_product_id(self):
         res = super()._onchange_product_id()
         if (not config['test_enable'] or
                 self.env.context.get('test_account_analytic_default_account')):
+            date = self.env['account.invoice'].browse(
+                self.invoice_id.id).date or fields.Date.today(),
             rec = self.env['account.analytic.default'].account_get(
                 product_id=self.product_id.id,
                 partner_id=self.invoice_id.partner_id.id,
-                user_id=self.env.uid, date=fields.Date.today(),
-                company_id=self.company_id.id, account_id=self.account_id.id
+                user_id=(self.invoice_id.user_id.id or self.env.uid),
+                date=date,
+                company_id=self.company_id.id,
+                account_id=self.account_id.id
             )
             self.account_analytic_id = rec.analytic_id.id
         return res
 
     def _set_additional_fields(self, invoice):
         if not self.account_analytic_id:
+            date = self.env['account.invoice'].browse(
+                invoice.id).date or fields.Date.today()
             rec = self.env['account.analytic.default'].account_get(
                 product_id=self.product_id.id,
                 partner_id=self.invoice_id.partner_id.id,
-                user_id=self.env.uid, date=fields.Date.today(),
-                company_id=self.company_id.id, account_id=self.account_id.id
+                user_id=self.env.uid,
+                date=date,
+                company_id=self.company_id.id,
+                account_id=self.account_id.id
             )
             if rec:
                 self.account_analytic_id = rec.analytic_id.id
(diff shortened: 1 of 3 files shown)


grindtildeath (commented):
This change seems a bit dubious to me as I don't understand why the default would be searched on another company. Isn't the module maybe missing some multi company `ir.rule`?
IMO would be better to begin with adding a unit test of what you described in this PR to highlight the issue.

i-vyshnevska:
@grindtildeath  by default it tries to do search with 
company=false | company=id

but when we searching  we call method with data company=false and domain becoming:
company=false | company=false 

to handle it appropriately we probably should replace company from context if it false,
but in this case we have real account  with defined company and can really this function on it, there never will be case to have account in one company and analytic_account in other

account_analytic_default_account/models/account_analytic_default_account.py
```diff
@@ -48,6 +48,10 @@ def account_get(self, product_id=None, partner_id=None, user_id=None,
         """Search the records matching the built domain,
         compute an index score based on the actual record values then
         return the record with the highest index score"""
+        # if we have account than company should be same as in account
```
asaunier:
```suggestion
        # if we have an account then the company should be the same as in the account
```

grindtildeath (commented):
Finally got the time to reproduce and analyse a bit more the issue.
IMO What must be changed are the calls to `account_get` here:
https://github.com/OCA/account-analytic/pull/237/files#diff-0c032965495b9af2dc678a57b3a4db33R118
https://github.com/OCA/account-analytic/pull/237/files#diff-0c032965495b9af2dc678a57b3a4db33R128
https://github.com/OCA/account-analytic/pull/237/files#diff-0c032965495b9af2dc678a57b3a4db33R136
Because that's also what's done in the std module:
https://github.com/odoo/odoo/blob/11.0/addons/account_analytic_default/models/account_analytic_default.py#L63
https://github.com/odoo/odoo/blob/11.0/addons/account_analytic_default/models/account_analytic_default.py#L71
And it's also done 'correctly' (i.e. with as many params as possible) in the overriden functions:
https://github.com/OCA/account-analytic/pull/237/files#diff-0c032965495b9af2dc678a57b3a4db33R93
https://github.com/OCA/account-analytic/pull/237/files#diff-0c032965495b9af2dc678a57b3a4db33R104

Since all the domain leaves will get `key = value or key = False` there shouldn't be any risk to add as many params as possible when calling `account_get`.

And IMO adding a test is still the best way to explicit such a change. :wink:

i-vyshnevska:
@grindtildeath  updated
cc @fclementic2c your review are welcome

grindtildeath (commented):
Looks better, but IMO these two calls to account_get must use more parameters as well:
 - https://github.com/OCA/account-analytic/pull/237/files#diff-0c032965495b9af2dc678a57b3a4db33R122
 - https://github.com/OCA/account-analytic/pull/237/files#diff-0c032965495b9af2dc678a57b3a4db33R131

sebalix:
Some lints remaining: https://travis-ci.org/OCA/account-analytic/jobs/538246338#L591

account_analytic_default_account/tests/test_analytic_default_account.py
```diff
@@ -141,3 +152,86 @@ def test_account_analytic_default_account_move(self):
         self.assertEqual(move_line_1.analytic_account_id,
                          self.analytic_account_4)
         self.assertFalse(move_line_2.analytic_account_id.id)
+
+        # set company
+        self.account_analytic_default_model.create({
+            'account_id': self.account_sales.id,
+            'analytic_id': self.analytic_account_1.id,
```
grindtildeath, line 163: IMO would be better to create one `account.analytic.account` for each `account.analytic.default` that you use in the test (as is done there https://github.com/OCA/account-analytic/pull/237/files#diff-467b4b47594859ebd20aea47ff40cf36R58 ), just to make sure the default record that is used is the one you created and not an existing one created in the `setUpClass`.

grindtildeath (commented):
Looks much better, thanks :smiley: 
Just one improvement in the test to avoid possible headaches.

account_analytic_default_account/tests/test_analytic_default_account.py
i-vyshnevska: @grindtildeath updated

grindtildeath (approved):
Thank you :hugs:

PR #223, opened 2019-01-31
[12.0][MIG] stock_analytic

AaronHForgeFlow:
supersedes #222

PR #242, opened 2019-06-13
[12.0][MIG] stock_analytic (updated the tests for the module stock_analytic)

BT-cserra:


pedrobaeza:
There's already #223 for this module. Please review that one instead.

PR #251, opened 2019-09-23
[12.0][ADD] account_tag_dimension_enhanced

kittiu:
This module enhance features for Analytic Tags based on Dimension, allowing,

* Set required dimension's tags
* Select tags by dimension sequence
* Mass create tags based on model data
* Filter following tag selections based on previous selected dimension's tags

With this, it is now possible to work as following, i.e.,
* We want 2 new dimension tags, 1) Department 2) Employee (just for testing)
* User must select Department first, which will filter only Employee belong to the department

### Setup

1. Create tags out of Department, set Sequence = 1

Selection_344

2. Create tags out of Employee, set Sequence = 2, and to be filtered by Department

Selection_345

### Usage

1. First selection, user will be entitled to select only departments (and not employee)

Selection_346

2. Second selection, user can easily select employee as it is filtered by 1st select

Selection_347


kittiu:
Thank you for your wonderful original module `account_tag_dimension`. I have long problem and looking for this kind of solution like yours. :)
I do have more requirements and so, came up with the enhancement. Do you prefer to have it separated or you think it is ok to be included in your module?

kittiu:
Out of this, I do have one resolved technical feature.
While, I am using api.onchange to return the dynamically computed `domain`. But the domain disappeared once I save it.
Are there way to preserve domain after on change?

fclementic2c:
@kittiu I am a big fan of this module too. Wonderful improvement!

kittiu:
Hello @fclementic2c @grindtildeath 

I am not sure who should I be talking to for this module, but I think this owned by c2c?

In fact, I like to propose as enhancement PR to the existing `analytic_tag_dimension` than to create a separated PR.

Mainly because, I will need to do @api.onchange('not common name field') on all underlining model.

https://github.com/ecosoft-odoo/account-analytic/blob/12.0-add-account_tag_dimension_enhanced/analytic_tag_dimension_enhanced/models/account_invoice.py#L10-L12

This means that, given there is module `analytic_tag_dimension_sale_warning` for sale.order.line, I would need to create another module, i.e., `analytic_tag_dimension_sale_warning_enhanced` just to add the onchange.

If I can enhanced this to `analytic_tag_dimension`, I could by the same, just to enhance `analytic_tag_dimension_sale_warning` and we don't end up with a lot of modules.

What do you think? Or who is the right person to talk to?

kittiu:
> @api.onchange('not common name field')

This problem is already been fixed, by not using it and use @api.depends instead.
cc @Saran440

kittiu:
This module is ready for review.

Saran440 (approved):
LGTM. Functional Test :+1:

kittiu:
@fclementic2c @grindtildeath I think the problem is all fixed. If you could take a look. Thanks!

rafaelbn (approved):
Thank you :+1:

PR #252, opened 2019-09-23
[11.0] Add account_analytic_salesperson

grindtildeath:
This module ensures that analytic accounts are created for each sale order that
is confirmed ands adds a Salesperson field on analytic accounts and analytic
lines in order to keep track of the Salesperson that was defined on the sale
order when the analytic account is created.


vrenaville (approved):

account_analytic_salesperson/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class SaleOrder(models.Model):
+
+    _inherit = 'sale.order'
+
+    @api.multi
+    def _create_analytic_account(self, prefix=None):
+        """Propagate sale order salesperson through context key"""
+        self = self.with_context(analytic_user_id=self.user_id.id)
```
rousseldenis: I think this is not correct.

This function is multi and will fail when wanting to access user_id field in multi recordset.

account_analytic_salesperson/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+        return super()._create_analytic_account(prefix=prefix)
+
+    @api.multi
+    def action_confirm(self):
+        """Ensure analytic account is created after sale order confirmation"""
+        res = super().action_confirm()
+        for order in self:
+            if not order.analytic_account_id:
```
rousseldenis, line 22: IMHO, even if you mention in DESCRIPTION that you ensure analytic account is created for each sale order, it is a change in the standard behaviour. Is it really the needs ?

You use the expense product here https://github.com/OCA/account-analytic/pull/252/files#diff-938c8d23d854c68b16b71b3ec1bcd046R12

So, the Odoo stnadard behaviour will apply, no ?

rousseldenis (changes requested):
Code review. Few questions. One requested change

account_analytic_salesperson/models/sale_order.py
grindtildeath: You're right, I thought I could do this after this fix https://github.com/odoo/odoo/commit/1377c45f4dd069b1fc38b8d0b922a22345470c20, but I should better add a loop in there.

account_analytic_salesperson/models/sale_order.py
grindtildeath: @rousseldenis Yes, it is intended (as stated in readme). 
I'll add a note in ROADMAP, that it could probably be split into two modules but at the moment I don't have time to rework it that much.

The expense product is only used in the test to make sure behaviour doesn't change :wink:

account_analytic_salesperson/models/sale_order.py
rousseldenis: Yeah but IMHO that PR is not correct as _create_analytic_account is multi. That one introduces a regression in terms of performance.

Should have been : self.filtered(lambda r: not r.analytic_account_id)._create_analytic_account()

grindtildeath: We cannot do that here because we need to pass the context for each record.

But anyway, this function is never called on a multi recordset in odoo:
 - https://github.com/odoo/odoo/blob/11.0/addons/sale/models/sale.py#L557
 - https://github.com/odoo/odoo/blob/11.0/addons/sale/models/sale.py#L873
 - https://github.com/odoo/odoo/blob/11.0/addons/sale_expense/models/hr_expense.py#L26
 - https://github.com/odoo/odoo/blob/11.0/addons/sale_timesheet/models/sale_order.py#L202

So it would actually be cleaner to have an `ensure_one` here https://github.com/odoo/odoo/blob/11.0/addons/sale/models/sale.py#L578 but as I see it, there's no real regression in terms of performance.

account_analytic_salesperson/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+This module ensures that analytic accounts are created for each sale order that
```
rafaelbn, line 1: IMHO this shouldn't be controlled by a module because to be sure you should have a correct configuration in the product or just inform the analytic account in the sale order

account_analytic_salesperson/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
...
+is confirmed ands adds a Salesperson field on analytic accounts and analytic
+lines in order to keep track of the Salesperson that was defined on the sale
+order when the analytic account is created.
```
rafaelbn, line 4: But the saleperson is already in the account.invoice , why to take it from the sale order and nor from the invoice?

account_analytic_salesperson/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+* As the module now includes two features (Creation of analytic account for
+  every sale order and addition of the salesperson field), this module could
+  be split into two different modules.
```
rafaelbn, line 3: I personally don't would make this solution. But of course I don't know the need. If you need that every sale order have analytic account just configure the product as create project, or just make a little customization to make the field analytic_account mandatory in the sale order :thinking:

account_analytic_salesperson/readme/DESCRIPTION.rst
yterrettaz: We would like to create in one operation the analytic account and the salesperson linked to this account. In fact the sales person will be the same from SO and from Invoice.

account_analytic_salesperson/readme/ROADMAP.rst
yterrettaz: Sometimes we need to create a analytic account even the sales order doesn't contain a product type service. Hardware for example. I can't use the standard to create a project if I sell a physical product.

account_analytic_salesperson/models/sale_order.py
```diff
@@ -0,0 +1,24 @@
...
+        for order in self:
+            order = order.with_context(analytic_user_id=order.user_id.id)
+            super(SaleOrder, order)._create_analytic_account(prefix=prefix)
+
+    @api.multi
+    def action_confirm(self):
+        """Ensure analytic account is created after sale order confirmation"""
+        res = super().action_confirm()
+        for order in self:
+            if not order.analytic_account_id:
```
i-vyshnevska: @grindtildeath 
can we have instead:
     
       for order in self:
            if not order.analytic_account_id:
                order._create_analytic_account()
        return super().action_confirm()

in your case analytic_account_id will not be propagated with context

account_analytic_salesperson/models/sale_order.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @api.multi
+    def _create_analytic_account(self, prefix=None):
+        """Propagate sale order salesperson through context key"""
+        for order in self:
+            order = order.with_context(analytic_user_id=order.user_id.id)
+            super(SaleOrder, order)._create_analytic_account(prefix=prefix)
+
+    @api.multi
+    def action_confirm(self):
```
i-vyshnevska:
```suggestion
    def _action_confirm(self):
```

account_analytic_salesperson/models/sale_order.py
```diff
@@ -0,0 +1,24 @@
...
+        """Ensure analytic account is created after sale order confirmation"""
+        res = super().action_confirm()
```
i-vyshnevska:
```suggestion
        res = super()._action_confirm()
```

account_analytic_salesperson/models/sale_order.py
grindtildeath: As seen together changing the override from `action_confirm` to `_action_confirm` fixes the issue. Thanks for the suggestion :+1:

PR #258, opened 2019-10-16
[12.0] [FIX] stock_analytic: Analytic account field is in wrong page

### LINKED ISSUE #257: [12.0] stock_analytic: analytic_account_id field is read only!
I went to a Odoo runbot for 12.0 and tried to do just as the README file says and the field is readonly:

Jerther:
Fixes #257

### DIFF
--- a/stock_analytic/views/stock_move_views.xml
+++ b/stock_analytic/views/stock_move_views.xml
@@ -66,7 +66,7 @@
             <field name="model">stock.picking</field>
             <field name="inherit_id" ref="stock.view_picking_form"/>
             <field name="arch" type="xml">
-                <xpath expr="//field[@name='move_line_ids_without_package']/tree/field[@name='product_id']" position="after">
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
:+1: LGTM

MiquelRForgeFlow (changes requested):
I think there should be **two** xpaths instead. One for the `move_line_ids_without_package` and the other for the `move_ids_without_package`.

Jerther:
@mreficent The problem is, the field is readonly in `move_line_ids_without_package` and pretty much useless (see related issue). Adding `readonly=False` to the field declaration works but the field becomes blank when saving so I'd have to dig deeper.

In fact, correct me if I'm wrong, but the fields used to have different names in 11.0 so could it be that the inherited view used the wrong field in the migration from 11.0 to 12.0? If so, this PR really is a fix to that problem and we could make a separate issue of adding the analytic account field to `move_line_ids_without_package`, which I'm actually willing to take care of eventually but not right now :wink:

MiquelRForgeFlow (approved):
ok

PR #264, opened 2019-10-30
[13.0][MIG] analytic_tag_dimension

Saran440:


kittiu:
@Saran440 are you working on this module. If so, please fix travis and I can review it.

Thanks!

kittiu:
Functional Tested, LGTM :+1:

kittiu (approved):

analytic_tag_dimension/models/analytic.py
```diff
@@ -0,0 +1,108 @@
...
+    analytic_tag_ids = fields.One2many(
+        comodel_name="account.analytic.tag",
+        inverse_name="analytic_dimension_id",
+        string="Analytic Tags",
+    )
+
+    @api.model
+    def create(self, values):
+        if " " in values.get("code"):
+            raise ValidationError(_("Code can't contain spaces!"))
```
i-vyshnevska: this should be in constrain

analytic_tag_dimension/models/analytic.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def create(self, values):
+        result = super(AnalyticDimensionLine, self).create(values)
+        for value in values:
+            if value.get(result._analytic_tag_field_name):
+                result._handle_analytic_dimension()
+        return result
+
+    def write(self, values):
+        result = super(AnalyticDimensionLine, self).write(values)
```
i-vyshnevska:
```suggestion
        result = super().write(values)
```

analytic_tag_dimension/models/analytic.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def _handle_analytic_dimension(self):
+        for adl in self:
+            tag_ids = adl[self._analytic_tag_field_name]
+            tag_ids._check_analytic_dimension()
+            dimension_values = tag_ids.get_dimension_values()
+            super(AnalyticDimensionLine, adl).write(dimension_values)
+
+    def create(self, values):
+        result = super(AnalyticDimensionLine, self).create(values)
```
i-vyshnevska:
```suggestion
        result = super().create(values)
```

analytic_tag_dimension/models/analytic.py
```diff
@@ -0,0 +1,108 @@
...
+                            "name": "x_dimension_{}".format(values.get("code")),
+                            "field_description": values.get("name"),
+                            "ttype": "many2one",
+                            "relation": "account.analytic.tag",
+                        },
+                    )
+                ]
+            }
+        )
+        return super().create(values)
```
i-vyshnevska: I would do a call to super in first and then create fields based on self

analytic_tag_dimension/models/analytic.py
```diff
@@ -0,0 +1,108 @@
...
+            tag_ids = adl[self._analytic_tag_field_name]
+            tag_ids._check_analytic_dimension()
+            dimension_values = tag_ids.get_dimension_values()
+            super(AnalyticDimensionLine, adl).write(dimension_values)
+
+    def create(self, values):
+        result = super(AnalyticDimensionLine, self).create(values)
+        for value in values:
+            if value.get(result._analytic_tag_field_name):
+                result._handle_analytic_dimension()
```
i-vyshnevska:
```suggestion
        if values.get(result._analytic_tag_field_name):
           result._handle_analytic_dimension()
```
you cannot iterate here, and actually don't need to

i-vyshnevska (changes requested):

analytic_tag_dimension/models/analytic.py
Saran440: @i-vyshnevska values are list in dict can't used get

analytic_tag_dimension/models/analytic.py
Saran440: @i-vyshnevska okay, I fixed but when create dimension code space between letters it showed error
`Field names can only contain characters, digits and underscores (up to 63).`

analytic_tag_dimension/models/analytic.py
i-vyshnevska: maybe you about create_multi decorator but in create method we have dict, not a list, current PR is broken 
https://github.com/OCA/account-analytic/blob/ee2001213f40bbdd6e9943a2d29045c6cf6b8e71/analytic_tag_dimension/models/analytic.py#L85

Saran440: OH. Sorry, It's my mistake that delete `@api.model` in function create. Fixed this file and this PR. Thank you for your advise :)

Saran440: @i-vyshnevska `@api.model` in v12 and v13 is difference, we can used `@api.model_create_multi` instead. However it's return dict in list, So I used `filter` function to check condition instead of for loops. What do you think?

i-vyshnevska: even if @api.model was changed we shouldn't fill any difference on this level so it should work with it as well
suppose you can use  model_create_multi 
will see on tests

Saran440: model_create_multi will return values is dict in list. So, at first I used for loop

pedrobaeza:
Please include #303

Saran440:
move on https://github.com/OCA/account-analytic/pull/307

PR #288, opened 2020-02-07
[12.0][FIX] stock_analytic: moves with different analytic accounts

Jerther:
Moves with different analytic accounts should not be merged.

This PR just adds `analytic_account_id` to the list of distinctive fields.


AaronHForgeFlow (approved):
Agree

ps-tubtim (approved):

PR #295, opened 2020-03-20
13.0 mig account analytic distribution required

jimhoefnagels:


account_analytic_distribution_required/models/account.py
```diff
@@ -0,0 +1,88 @@
...
+    @api.constrains(
+        "analytic_account_id", "analytic_tag_ids", "account_id", "debit", "credit"
+    )
+    def _check_analytic_required(self):
+        for rec in self:
+            message = rec._check_analytic_distribution_required_msg()
+            if message:
+                raise exceptions.ValidationError(message)
+            else:
+                super(AccountMoveLine, self)._check_analytic_required()
```
rven:
```suggestion
    def _check_analytic_required(self):
        for rec in self:
            message = rec._check_analytic_distribution_required_msg()
            if message:
                raise exceptions.ValidationError(message)
        super()._check_analytic_required()
```
No need to check all records for each record, all at once after this check is enough

rven (changes requested):

account_analytic_distribution_required/tests/test_account_analytic_plan_required.py
```diff
@@ -0,0 +1,182 @@
...
+from datetime import datetime
+
+from odoo.exceptions import ValidationError
+from odoo.tests.common import SavepointCase
+
+
+class TestAccountAnalyticPlanRequired(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestAccountAnalyticPlanRequired, cls).setUpClass()
```
rven:
```suggestion
        super().setUpClass()
```

PR #303, opened 2020-05-16
[12.0][IMP] analytic_tag_dimension: Sync dimension code and description changes

carlosdauden:
Syncronice changes in exsiting dimensions.
Allow extend models.

@Tecnativa

### DIFF
--- a/analytic_tag_dimension/models/analytic.py
+++ b/analytic_tag_dimension/models/analytic.py
@@ -1,5 +1,6 @@
 # Copyright 2017 PESOL (http://pesol.es)
 #                Angel Moya (angel.moya@pesol.es)
+# Copyright 2020 Tecnativa - Carlos Dauden
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import models, api, fields, _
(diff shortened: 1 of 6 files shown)


analytic_tag_dimension/models/analytic.py
```diff
@@ -18,28 +19,51 @@ class AccountAnalyticDimension(models.Model):
...
-        )
+        ]
+
+    def get_field_name(self, code=False):
+        return 'x_dimension_{}'.format(code or self.code)
+
+    @api.model
+    def create(self, values):
+        if ' ' in values.get('code'):
+            raise ValidationError(_("Code can't contain spaces!"))
```
pedrobaeza: Why not putting this as `api.constrains`?

pedrobaeza (commented):
Now that I'm seeing the code, I think it would be desirable to inject the fields the same way we do in `base_search_field_filter`, don't you think?

analytic_tag_dimension/models/analytic.py
```diff
@@ -18,28 +19,51 @@ class AccountAnalyticDimension(models.Model):
         string='Analytic Tags')
 
     @api.model
-    def create(self, values):
-        if ' ' in values.get('code'):
-            raise ValidationError(_("Code can't contain spaces!"))
-        model_names = (
+    def get_model_names(self):
```
kittiu, line 28: Good enhancement. I have one question. As this module is used to extend dimension in other module, i..e, sale.order.line, purchase.order.line
https://github.com/ecosoft-odoo/account-analytic/blob/12.0-mig-analytic_tag_dimension_sale_warning/analytic_tag_dimension_sale_warning

Will this PR has affect on those?

rafaelbn (approved):
Functionally tested :+1: 

Thanks

angelmoya (approved):

pedrobaeza:
@carlosdauden what do you think about my proposal?

carlosdauden:
@pedrobaeza I don't know if I understand what you mean.
If it is to create virtual fields to be able to filter, but without the possibility of grouping, I think that precisely the main objective of this module is the possibility of grouping in the pivot views.

pedrobaeza:
Well, I mean to not require the ir.model.fields for doing all (filtering, grouping), injecting them on the fly and relying in a computed field. Is it possible?

carlosdauden:
I don't know

analytic_tag_dimension/models/analytic.py
carlosdauden: I think that without this PR, the linked module don't have a inheritable way to create dimension columns in mentioned models, if you need.
So this PR beneficially affects if you extend `get_model_names` and does not affect at all if you don't.

pedrobaeza:
OK, let's go as is and let the possible implementation for further investigations:

/ocabot merge minor

PR #300, opened 2020-03-29
[13.0][MIG] analytic_tag_dimension_sale_warning

i-vyshnevska:
migrate analytic_tag_dimension_sale_warning
based on #280
build requires merge #264

PR #299, opened 2020-03-28
[13.0][MIG] analytic_tag_dimension_purchase_warning

i-vyshnevska:
migrate analytic_tag_dimension_purchase_warning
based on https://github.com/OCA/account-analytic/pull/246
build requires merge https://github.com/OCA/account-analytic/pull/264

PR #312, opened 2020-07-24
[13.0][mig] analytic_tag_dimension_sale_warning

captivea-dcr:
superseeds #300


baimont:
Functional testing: ko

To reproduce:
The warning message is there, so the module works. But it misses something.
image

Problem is, when you save above sale order, you get:
ValueError: Invalid field 'x_dimension_type' on model 'sale.order.line'

What is missing here is an inheritance of `get_model_names()` method on `account.analytic.dimension` in order to append `sale.order.line` to model names. This method is called whenever you create or write on a dimension to create or update the field in said model.

I haven't checked, but the problem should arise too for
https://github.com/OCA/account-analytic/pull/313
https://github.com/OCA/account-analytic/pull/300
https://github.com/OCA/account-analytic/pull/299

baimont (approved):
functional testing ok

kittiu (approved):
LGTM

PR #313, opened 2020-07-24
[13.0][mig] analytic_tag_dimension_purchase_warning

captivea-dcr:
superseeds #299


kittiu:
Base off your branch, I got error upon following test

1. New demo database
2. Install this module
3. Setup to open up analytic tag field.
4. Create a PO, and choose Analytic Tag = 'Dim Type A' or 'Concept A' and save. (no error if I choose 'Contracts')

```
    super(AnalyticDimensionLine, adl).write(dimension_values)
  File "/opt/odoo/auto/addons/purchase/models/purchase.py", line 624, in write
    return super(PurchaseOrderLine, self).write(values)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3536, in write
    raise ValueError("Invalid field %r on model %r" % (fname, self._name))
ValueError: Invalid field 'x_dimension_type' on model 'purchase.order.line'
```

kittiu:
It looks link your branch code is not updated, see this.
https://github.com/OCA/account-analytic/tree/13.0/analytic_tag_dimension/models

Can you rebase?

kittiu:
Looks like this simple module can have error. May be adding test is a good idea. :)

PR #323, opened 2020-09-30
[IMP] 11.0 - Account Analytic Parent: add complete name

j-zaballa:
* add: `complete_name` computed field
* add: override `name_search` so we can search by complete name
* imp: `name_get` backport from v12 and remove @api.depends decorator
* imp: search view - add complete name to name's filter_domain
* ref: move `_check_parent_active` so the structure is compliant with OCA guidelines
* ref: remove `super` arguments when possible


j-zaballa:
I believe Travis is red because of this:

https://github.com/OCA/account-analytic/blob/021f85f582081a75abcd99ede218f46eaf7ab8dc/account_analytic_parent/i18n/fr.po#L48

I can't translate into french on Weblate though...

PR #328, opened 2020-10-26
[12.0][IMP][product_analytic] - get income/expense analytic account/tag from the nearest parent category

sbejaoui:
1. add default analytic tag to the product and category models
2. when the account/tag is not defined on the product and the category, get the nearest parent category with an account/tag


sbejaoui:
cc/ @gva-acsone

PR #329, opened 2020-10-27
[12.0][ADD] - new module crm_team_analytic_tag

sbejaoui:
This addon add analytic tags to sale teams to be added to invoice lines when the team is selected in the invoice level


sbejaoui:
cc/ @gva-acsone

PR #330, opened 2020-10-27
[12.0][IMP+FIX+REF] analytic_tag_dimension: Refactor + performance imp + fix on removal + handle dimension removal / module uninstallation

pedrobaeza:
This module required a need of love. I have performed some refactoring for being more maintainable, with a performance improvement not writing 2 times on records, and also a fix when removing tags. These are the commits messages:

* [REF] Follow OCA guidelines of one file per main model
* [REF] Refactor tests
  - More simplicity
  - Don't use demo data
  - Test all conditions on each model
* [IMP] Don't write 2 times when using tags

  Previous code was doing a deferred write after the creation/first write for easing the code, but it's very bad for performance in such used objects.

  This commits changes the way this is handled for covering the cases in advance, and thus not requiring a second write. The only drawback is that writing on multiple records means to perform a write on each record separately. This case was also happening
previously, and we have gained one write (the initial one), but it's not the ideal one in terms of performance. A possible improvement can be to group the records with the same final values to be performed at once.
* [FIX] Handle removal tags

  When you remove tags, dimensions fields were not updated accordingly before this commit.
* [FIX] Handle dimension removal / module uninstallation

  If not, there is remaining garbage in ir.model.fields and SQL columns in tables, avoiding a second install of the module or to reuse the same dimension code.

@Tecnativa TT26481


joao-p-marques (approved):

pedrobaeza:
I have added more changes, as this module lets garbage on dimension removal or module uninstallation. I also overpass the limitation of using commands 0 and 1 on tags. Finally, all these changes deserves a promotion to co-author IMO. If you don't agree, please tell.

analytic_tag_dimension/models/account_analytic_tag.py
```diff
@@ -0,0 +1,62 @@
...
+        if 'analytic_dimension_id' in vals:
+            Dimension = self.env['account.analytic.dimension']
+            _models = [self.env[m] for m in Dimension.get_model_names()]
+            for tag in self.filtered('analytic_dimension_id'):
+                old_field = tag.analytic_dimension_id.get_field_name()
+                new_field = Dimension.browse(
+                    vals['analytic_dimension_id']).get_field_name()
+                if old_field == new_field:  # pragma: no cover
+                    continue
+                # Filter to avoid update report models
```
victoralmau, line 42:
```suggestion
            for tag in self.filtered('analytic_dimension_id'):
                Dimension = self.env['account.analytic.dimension']
                old_field = tag.analytic_dimension_id.get_field_name()
                new_field = Dimension.browse(
                    vals['analytic_dimension_id']).get_field_name()
                if old_field == new_field:  # pragma: no cover
                    continue
                # Filter to avoid update report models
                 _models = [self.env[m] for m in Dimension.get_model_names()]
```

analytic_tag_dimension/models/analytic_dimension_line.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @api.multi
+    def write(self, vals):
+        """Inject values for dimension fields."""
+        if self._analytic_tag_field_name in vals:
+            for record in self:
+                vals = record._handle_analytic_dimension(vals)
+                super(AnalyticDimensionLine, record).write(vals)
+            return True
+        return super().write(vals)
```
victoralmau, line 74:
```suggestion
"""Inject values for dimension fields."""
        new_vals_list = []
        if self._analytic_tag_field_name in vals:
            for record in self:
                new_vals_list.append(record._handle_analytic_dimension(vals))
        else:
            new_vals_list = vals
        return super().write(new_vals_list)
```

analytic_tag_dimension/models/account_analytic_dimension.py
```diff
@@ -0,0 +1,84 @@
...
+    def create(self, values):
+        res = super().create(values)
+        _models = self.env['ir.model'].search([
+            ('model', 'in', self.get_model_names()),
+        ])
+        _models.write({
+            'field_id': [(0, 0, {
+                'name': self.get_field_name(values['code']),
+                'field_description': values.get('name'),
+                'ttype': 'many2one',
```
victoralmau, line 48: `ttype` is correct field?  It seems incorrect (`type`)

victoralmau (approved):

analytic_tag_dimension/models/account_analytic_tag.py
pedrobaeza: I don't get why...

analytic_tag_dimension/models/analytic_dimension_line.py
pedrobaeza: Uhm, I put that... not sure if I have lost that changes...

analytic_tag_dimension/models/account_analytic_dimension.py
pedrobaeza: Yes, it's correct.

analytic_tag_dimension/models/analytic_dimension_line.py
pedrobaeza: Ah, no, this one is not correct. This is not a creation, but a write, and you can't pass multiple update dictionaries to super. You can only call super record per record with the proper update dictionary. Please read commit messages with further explanations.

analytic_tag_dimension/models/account_analytic_tag.py
victoralmau: Only small details to improve code because `Dimension = self.env['account.analytic.dimension']` not need to load before filtered, and `_models = [self.env[m] for m in Dimension.get_model_names()]` similar.

pedrobaeza: It's done for performance. Changing it outside the loop, you avoid to call it N times.

analytic_tag_dimension/models/analytic_dimension_line.py
victoralmau: Only little suggestion to use `super()`  only 1 time (and not in every record)

pedrobaeza: But that's not possible as stated. Please read again my previous comment and also the commit messages, and then tell me if you understand it.

joao-p-marques (approved):

sbidoul:
@sbejaoui @gva-acsone

rafaelbn (approved):
:+1:

pedrobaeza:
Cherry-picked for v13 in:

5e2e56d809cc519735216c0d140e1157da32e1c2
b161abdf8e225b5ac3b75cdb1f038ca9aa1d19ec
a83f6808eaaddf54d36a6fadd5fd8e9239931eb4
104f24a15e43754238122516ba1051db47c3a9e7
a944229a962633cec40b751493b2cbedf32a147f
cb7917e1245325006ab603bbadc04a89e558e033

PR #335, opened 2020-11-11
[13.0][ADD] new module stock_landed_costs_analytic

baimont:
New module that copies the analytic account and the analytic tags of the landed cost lines to the journal items


apineux (approved):

sanrav (approved):
Functional testing OK

image

image

PR #338, opened 2020-11-26
[13.0][IMP] analytic_base_department - groupby department and select analytic account from department

Saran440:
Implement module analytic_base_department following:
- Groupby department on analytic account
Selection_005

- Add new feature, helping user add analytic account from department
Selection_004
change it to table
Selection_001

- Add summary module on manifest

PR #339, opened 2020-11-26
[14.0][MIG] analytic_base_department

Saran440:
Standard Migrate and cherry-pick from https://github.com/OCA/account-analytic/pull/338

@rafaelbn I think this module still used because
- Module hr_timesheet is not necessary install for all project
- There is field `department_id` on `account_analytic_line` when install module `hr_timesheet` BUT it not display.
- In core odoo, `account_analytic` without department_id


analytic_base_department/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2011-2018 Camptocamp SA
+# Copyright 2013 Daniel Reis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Base Analytic Department Categorization",
```
MeetKD, line 5: @Saran440 It would be great if you can add summary about the module !

Saran440: Okay I will. Thank you for your reviewed. :)

gurneyalex (approved):

analytic_base_department/__manifest__.py
Saran440: @MeetKD I added it. Please, Could you review and approved this?

PR #342, opened 2020-11-26
[FIX][13.0] fix travis after merge (instead of before merge)

baimont:


pedrobaeza:
That's definitively shouldn't be there in travis file at all

/ocabot merge nobump

PR #345, opened 2020-12-16
[12.0][FIX] Remove unnecessary translated term that is error prone

ypapouin:
No translatable content


ypapouin:
Didn't think about RTL languages

PR #348, opened 2020-12-31
[14.0] [ADD] analytic_activity_based_cost: New module to add a cost line from my analytic lines.

### LINKED ISSUE #347: [14.0][RFC] analytic_activity_based_cost: Anaytic Items for Activity Based Costing
# Description

Generate Activity Based Costs (ABC), using Analytic Items. 

Activity Based Costing is a well documented technique to assign indirect costs to the the activities identified as cost drivers.
This feature models cost driver usage as Analytic Items.
When an Analytic Item is created, it may then generate additional Analytic Items for the corresponding indirect costs.
For example, each timesheet hour logged could generate a quantity and amount of overhead assigned to that activity.

This feature does not generate Account Moves.
That should be implemented by a separate feature.

# Usage

When creating Analytic Items, if a configuration is in place, the corresponding Analytic Items for indirect cost are generated.

- When an Analytic Item is created, an automatic process checks the Activity Based Cost Rules to identify the ones that apply.
- Each triggered rule created a new Analytic Item, with a copy of the original one, and:
  - Product: is the rule Cost Type Product. A validation error prevents this from being the same as the source Analytic Item Product, to avoid infinite loops. 
  - Quantity: is the original quantity multiplied by the rule's Factor
  - Amount: is -1 * Quantity  * Product Standard Price
  - Parent Analytic Item (new field): set with the original Analytic Item
- An update on the Quantity triggers a recalculation of the quantity and amount of the child Analytic Items.
- A delete cascades to the child Analytic Items, causing them to also be deleted.
 
# Configuration

Create the Products representing the Cost Types to use:

- Go to Invoicing/Accounting > ... > Products, and create Products representing the Cost Types to use. Set the standard cost to use for each unit used.

Configure the Activity Based Cost Rules, setting the triggers to generate the cost Analytic Items:

- Go to Invoicing/Accounting > Configuration > Activity Based Cost Rules.
- Create a new rule:
  - Name: a descriptive title, required
  - Start Date: an optional application start date
  - End Date: an optional application end date
  - Activity Product: the Analytic Item Product generating costs, if given. Optional.
  - Must have related Project: check to only apply if the Analytic Item is related to a Project. Allows to identify timesheet generated Analytic Items.
  - Factor: a quantity multiplier to calculate the cost type quantity. Defaults to 1.
  - Cost Type Product: a Product that represents the cost type to assign, such as "Overhead". The Product's standard cost will be used to calculate amounts.

Chandresh-SerpentCS:
Closes #347


dreispt (approved):
Made review changes and pushed them directly to the branch.

PR #352, opened 2021-01-21
[13.0][ADD] procurement_mto_analytic - only add account_analytic_id in po domain if value is there

baimont:
This pr adapts method override _make_po_get_domain in procurement_mto_analytic.

Running scheduler to create purchase orders before this pr would create one purchase order per reordering rule. We do not want that.

### DIFF
--- a/procurement_mto_analytic/models/stock_rule.py
+++ b/procurement_mto_analytic/models/stock_rule.py
@@ -17,11 +17,12 @@ class StockRule(models.Model):
 
     def _make_po_get_domain(self, company_id, values, partner):
         res = super()._make_po_get_domain(company_id, values, partner)
-        res += (
-            (
-                "order_line.account_analytic_id",
-                "=",
-                values.get("account_analytic_id", False),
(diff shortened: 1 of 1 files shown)


procurement_mto_analytic/models/stock_rule.py
```diff
@@ -17,11 +17,12 @@ def _prepare_purchase_order_line(
...
     def _make_po_get_domain(self, company_id, values, partner):
         res = super()._make_po_get_domain(company_id, values, partner)
-        res += (
-            (
-                "order_line.account_analytic_id",
-                "=",
-                values.get("account_analytic_id", False),
-            ),
-        )
+        if "account_analytic_id" in values and values["account_analytic_id"]:
```
alan196, line 20:
```suggestion
        if values.get("account_analytic_id", False):
```

alan196 (approved):
👍  LGTM, Technical review

adrienpeiffer (approved):
LGTM

PR #357, opened 2021-02-11
[13.0] [MIG] account_analytic_no_lines from 12.0

ThomasBinsfeld:
From https://github.com/OCA/account-analytic/pull/191
#253


Laurent-Corron (approved):

account_analytic_no_lines/models/account_move_line.py
```diff
@@ -0,0 +1,23 @@
...
+from odoo import api, models
+
+
+class AccountMoveLine(models.Model):
+
+    _inherit = "account.move.line"
+
+    @api.model
+    def _must_create_analytic_line(self, move_line):
+        if not move_line.product_id:
```
etobella, line 13: Why do you want to do this if you are hidding it? Shouldn't be False by default? Why using expense_policy? In this case, maybe it has sense to add it somewhere else, because expense_policy is defined on `sale`.... WDYT?

PR #363, opened 2021-02-18
14.0 mig account analytic parent

cvinh:


rvalyi (approved):
LGTM.

However, it's important to bare in mind that in recent Odoo versions, you should structure the analytic accounts just like the general accounts, using an analytic account group hierarchy. So if one goes acccount_analytic_parent, he may miss support from the core or even from mis-builder.

May be worth a note in the README. At least I did this mistake on a v12 project,..

dreispt (approved):

PR #369, opened 2021-04-07
[10.0][IMP] analytic-partner: set other partner on existing analytic lines

pierre-halleux:
Set the other_partner field value on existing analytic lines.


analytic_partner/hooks.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Set the `other_partner` field on all existing analytic lines.
+    :param cr: database cursor
+    :param version: str
+    :return: None
+    """
+    _logger.info("Update other_partner field on existing analytic lines.")
+    env = Environment(cr, SUPERUSER_ID, {})
+
+    account_move_lines = env["account.move.line"].search([])
```
dreispt, line 22:
```suggestion
    account_move_lines = env["account.move.line"].search([("other_partner", "=", False)])
```

Issue #370, opened 2021-04-26
[13.0] account_analytic_required: faling tests on main branch


chienandalu:
13.0 tests failing in main branch https://travis-ci.com/github/OCA/account-analytic/jobs/500780750 (probably due to some upstream change)

PR #371, opened 2021-04-26
[FIX] account_analytic_required: fix test

chienandalu:
We need to rollback after the exception, otherwise the move stays posted
and the subsequent test isn't propperly evaluated

Solves https://github.com/OCA/account-analytic/issues/370

cc @Tecnativa TT27217


chienandalu:
@victoralmau please review

victoralmau (approved):

pedrobaeza (approved):

PR #377, opened 2021-05-17
[13.0][IMP] stock_analytic: Prevent error when set real_time valuation and stock.valuation.layer is auto-created 

victoralmau:
Prevent error when set **real_time** valuation and `stock.valuation.layer` is auto-created  (according to (according to products in test category).

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT29728

### DIFF
--- a/stock_analytic/tests/test_stock_picking.py
+++ b/stock_analytic/tests/test_stock_picking.py
@@ -3,19 +3,16 @@
 # Copyright 2016 OpenSynergy Indonesia
 # Copyright 2017 ForgeFlow S.L.
 # Copyright 2018 Hibou Corp.
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from datetime import datetime
-
-from odoo.tests.common import TransactionCase
+from odoo.tests import Form, common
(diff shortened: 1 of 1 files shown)


joao-p-marques (approved):

pedrobaeza (commented):
Can you please remove totally demo data for avoiding more side effects?

victoralmau:
Changes done.

stock_analytic/tests/test_stock_picking.py
```diff
@@ -59,7 +56,7 @@ def setUp(self):
         self.outgoing_picking_type = self.env.ref("stock.picking_type_out")
         self.incoming_picking_type = self.env.ref("stock.picking_type_in")
 
-        self.product_categ.update(
+        self.product_categ = self.env.ref("product.product_category_5").copy(
```
pedrobaeza: This is still using demo data as "template", so the side effects are the same: any modification on such demo data will affect here. Please create the product and the category from scratch.

victoralmau: Ok sorry, changes done.

joao-p-marques:
@pedrobaeza seems like the bot got stuck... Maybe reissue the command?

PR #379, opened 2021-05-18
[13.0][IMP] analytic_tag_dimension: Improve _reflect function according to prevent error if we use x_dimension_xxx field

victoralmau:
Improve `_reflect` function according to prevent error if we use `x_dimension_xxx` field in inherit models from: 
- account.move.line
- account.analytic.line
- account.invoice.report

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT29728

### DIFF
--- a/analytic_tag_dimension/models/analytic_dimension_line.py
+++ b/analytic_tag_dimension/models/analytic_dimension_line.py
@@ -1,5 +1,6 @@
 # Copyright 2017 PESOL (http://pesol.es) - Angel Moya (angel.moya@pesol.es)
 # Copyright 2020 Tecnativa - Carlos Dauden
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, models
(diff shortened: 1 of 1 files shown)


analytic_tag_dimension/models/analytic_dimension_line.py
```diff
@@ -10,6 +11,32 @@ class AnalyticDimensionLine(models.AbstractModel):
...
+            "account.move.line",
+            "account.analytic.line",
+            "account.invoice.report",
+        ]
+
+    def _reflect(self):
+        super()._reflect()
+        if (
+            self._inherit != self._name
+            and self._inherit in self.get_model_names()
```
pedrobaeza: Why are you limiting to these models? Any children model should have it, isn't it?

analytic_tag_dimension/models/analytic_dimension_line.py
```diff
@@ -10,6 +11,32 @@ class AnalyticDimensionLine(models.AbstractModel):
...
+            and self.env.context.get("check_fields", True)
+        ):
+            fields_inherit = self.env["ir.model.fields"].search(
+                [
+                    ("model_id.model", "=", self._inherit),
+                    ("name", "not in", self.fields_get_keys()),
+                ]
+            )
+            if fields_inherit:
+                model = self.env["ir.model"].search([("model", "=", self._name)])
```
pedrobaeza: Why do you need to search again the model if you already have it on any of the fields_inherit records?

analytic_tag_dimension/models/analytic_dimension_line.py
```diff
@@ -10,6 +11,32 @@ class AnalyticDimensionLine(models.AbstractModel):
...
+            for field in fields_inherit:
+                field.with_context(check_fields=False).copy({"model_id": model.id})
```
pedrobaeza: What this exactly does? In the ORM, this is creating a copy of the ir.model.fields record, so not sure this is correct.

victoralmau: It's necessary and create `ir.model.field` records lost.
When dimension record is created, auto-create "custom fields" in some models according to: https://github.com/OCA/account-analytic/blob/13.0/analytic_tag_dimension/models/account_analytic_dimension.py#L38-L59 but these process have side-efects but now work fine.

Example:
- `account.move.line` (inherit from `analytic.dimension.line`) > Before, fields are auto-created in these model.
- `custom.wizard.model` (inherit from `account.move.line`) > Now, fields are auto-created in these model too.

analytic_tag_dimension/models/analytic_dimension_line.py
victoralmau: https://github.com/OCA/account-analytic/pull/379#discussion_r634997336

pedrobaeza: I still don't understand how copying a field is the solution. This duplicates the record.

pedrobaeza (approved):
Checked internally and see the steps to reproduce the problem, that are now in commit message.

victoralmau:
I have added some little changes. 
Now it's ready to merge.

PR #271, opened 2020-01-07
[13.0][MIG] account_analytic_sequence

ps-tubtim:
Standard migration.

PR #383, opened 2021-05-28
[14.0] Migration account_analytic_sequence

petrus-v:
migrated from v12.0


petrus-v:
This new constraint is breaking `account_analytic_parent` unittest  is it the way to go to fix that in the same PR ? or is there a way to tell two modules are incompatible in the same repository ?

petrus-v:
> This new constraint is breaking `account_analytic_parent` unittest is it the way to go to fix that in the same PR ? or is there a way to tell two modules are incompatible in the same repository ?

I've cherry-picked travis config from @ps-tubtim (thanks) used in #271 waiting CI results !

ps-tubtim (approved):
LGTM

manuelregidor (approved):
LGTM

PR #397, opened 2021-09-06
[IMP] analytic_activity_based_cost: move all ABC fields to Product Variant, for better usability

dreispt:


dreispt:
It was confusing to open the Product Template form, see the Driven Cost checkbox, but not see the ABC tab.
So that checkbox was moved to Product Variant.

Additionally, the noupdate="1" was removed from demo data, as there is little point in it, and later modification of demo data becomes simpler.

cc @patrickrwilson

patrickrwilson (approved):
LGTM

PR #408, opened 2021-10-21
[FIX] analytic_activity_based_cost: rounding issue

RLeeOSI:
Fixes rounding issue in journal items.

Previous behavior: product cost was calculated by sum of raw materials and activity costs multiplied by qty factor, then rounded. Journal items were calculated by each cost multiplied by a qty factor, then rounded individually. The sums do not always match.

Example:  A finished good product uses $0.87 in raw materials, labor is $31.66/hr and takes 15min, overhead is $44.66/hr and takes 15min. When calculating the cost of the finished good from the BoM, it calculates as $0.87 + $31.66/4 + $44.66/4 = $19.95. When the clearing lines are calculated, it gets $0.87 for raw materials, $31.66/4 = $7.915 ≅ $7.92 for labor, and $44.66/4 = $11.165 ≅ $11.17 for overhead. 0.87 + 7.92 + 11.17 = $19.96 which gives a $0.01 rounding error.

Solution: Round each activity cost before summing them into the finished good cost.

### DIFF
--- a/analytic_activity_based_cost/models/product.py
+++ b/analytic_activity_based_cost/models/product.py
@@ -3,6 +3,7 @@
 
 
 from odoo import _, api, exceptions, fields, models
+from odoo.tools.float_utils import float_round
 
 
 class Product(models.Model):
@@ -31,6 +32,9 @@ class Product(models.Model):
         for product in self.filtered("is_cost_type"):
             if product.type in ["consu", "service"]:
                 product.standard_price = sum(
-                    x.standard_price * x.factor
+                    float_round(
+                        x.standard_price * x.factor,
+                        precision_rounding=product.currency_id.rounding,
+                    )
                     for x in product.sudo().activity_cost_ids
                 )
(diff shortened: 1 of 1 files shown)


RLeeOSI:
@dreispt  Please review

analytic_activity_based_cost/models/product.py
```diff
@@ -31,6 +32,9 @@ def onchange_for_standard_price(self):
         for product in self.filtered("is_cost_type"):
             if product.type in ["consu", "service"]:
                 product.standard_price = sum(
-                    x.standard_price * x.factor
+                    float_round(
+                        x.standard_price * x.factor,
+                        precision_rounding=product.currency_id.rounding,
```
dreispt: Prices can have more decimals than the used by the currency.
So it makes sense for costs to also be able to support more decimals.
IMO we should use the product price precision here.

RLeeOSI:
@dreispt  can you review again? This fix should still be needed to avoid rounding errors

dreispt (approved):

PR #422, opened 2021-12-22
[14.0][MIG] account_analytic_no_lines

alexis-via:


account_analytic_no_lines/models/account_move_line.py
```diff
@@ -0,0 +1,22 @@
...
+
+from odoo import api, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    @api.model
+    def _must_create_analytic_line(self, move_line):
+        if not move_line.product_id:
```
etobella, line 12: The same comments apply from #357

Expense_policy is defined on `sale`, so it should be False always, if you need that, that should go to another module. Isn't it?

AaronHForgeFlow: I agree, that would go to another module I think.

alexis-via: Of course, we could have yet-another glue module... but we have quite a lot already. That's why "hasattr" is used ; that way, it works both when sale is installed and when sale isn't installed, without having to maintain a separate module. The idea is to generate the analytic line in DB ONLY when we really need it, i.e. when we have a product with an expense policy.

rousseldenis:
@alexis-via Is this still applicable ? If yes, reopen it ? @rafaelbn

rousseldenis:
I'll do 16.0 migration

rafaelbn:
Applicable for all that need this! 😄

rafaelbn:
@AaronHForgeFlow @etobella @alexis-via @rousseldenis last review and try to merge it?

rousseldenis (approved):
Code review

etobella (approved):

PR #432, opened 2022-01-21
[13.0][IMP] account_analtyic_parent - change action cost/revenue domain to child_of

amcor:
Change to see with the button of the parent analyci line all the child's items.
Now, the compute of total balance include all childs items, but when you use the button in parent account, only show the direct items.

### DIFF
--- a/account_analytic_parent/__manifest__.py
+++ b/account_analytic_parent/__manifest__.py
@@ -12,7 +12,7 @@
     "name": "Account Analytic Parent",
     "summary": """
         This module reintroduces the hierarchy to the analytic accounts.""",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
This is changing the way the module is conceived, so unless explanations of why doing this, I don't see it proper.

amcor:
Hi @pedrobaeza , the change is to see with the button of the parent analyci line all the child's items.
Now, the compute of total balance include all childs items, but when you use the button in parent account, only show the direct items.  For me is weird, but if isnt proper i'll close the pr

pedrobaeza:
I let other to give their opinion to see what they prefer.

PR #433, opened 2022-01-27
[14.0][Fix_Improve]: account_analytic_required translation template and fr.po

smussie:
Fix translation that were not loading properly, here before:
Screenshot from 2022-01-20 11-48-45
After:
Screenshot from 2022-01-27 15-46-44

### DIFF
--- a/account_analytic_required/i18n/account_analytic_required.pot
+++ b/account_analytic_required/i18n/account_analytic_required.pot
@@ -4,8 +4,10 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 14.0\n"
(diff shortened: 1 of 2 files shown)


leemannd (approved):

santostelmo (approved):

i-vyshnevska (changes requested):
translations to merged addons is added/changed through weblate https://translation.odoo-community.org/
changes through git may cause conflicts, sorry but not LG

leemannd:
@i-vyshnevska In this case, the standard website does not seems to be a good solution because the `pot` file is broken.

i-vyshnevska (approved):

PR #438, opened 2022-02-14
[13.0][MIG][WIP] account_analytic_default_purchase: Migration to v13.0

sergio-teruel:
cc @Tecnativa


sergio-teruel:
Not needed
Include here https://github.com/odoo/odoo/blob/928ff1c6c647d1fdc7f54b51bb39ccfdd04d9c50/addons/account_analytic_default_purchase/models/purchase_order_line.py#L10-L37

PR #439, opened 2022-02-15
[13.0][FIX] account_analytic_required: Take into account if account move line has an analytic tag with analytic distribution

sergio-teruel:
cc @Tecnativa

ping  @carlosdauden @chienandalu

### DIFF
--- a/account_analytic_required/models/account.py
+++ b/account_analytic_required/models/account.py
@@ -56,6 +56,14 @@ class AccountMoveLine(models.Model):
             force_company=account.company_id.id
         ).property_analytic_policy
 
+    def _has_analytic_distribution(self):
+        # If the move line has an analytic tag with distribution, the field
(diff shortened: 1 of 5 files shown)


carlosdauden (approved):

chienandalu (approved):

account_analytic_required/models/account.py
```diff
@@ -56,6 +56,20 @@ def _get_analytic_policy(self, account):
...
+        # check
+        tags_with_analytic_distribution = self.analytic_tag_ids.filtered(
+            "active_analytic_distribution"
+        )
+        if (
+            tags_with_analytic_distribution
+            and tags_with_analytic_distribution.analytic_distribution_ids
+        ):
+            return True
+        return False
```
pedrobaeza:
```suggestion
        return bool(
            tags_with_analytic_distribution.analytic_distribution_ids
        )
```

PR #440, opened 2022-02-22
[14][FIX] account_analytic_required: Take into account if account move line has an analytic tag with analytic distribution

sergio-teruel:
cc @Tecnativa

ping @carlosdauden @chienandalu 

Forward port from v13 https://github.com/OCA/account-analytic/pull/439/commits/af28420a8d84bf5f0d5917b414c2c892d029fc31


carlosdauden (approved):

rousseldenis (approved):
Code review. Seems legit

PR #447, opened 2022-04-01
[FIX] Check account analytic on PO line candidate search.

adrienpeiffer:
Currently we search on SO analytic account during the PO candidate search phase.
If there is a PO for the selected supplier but there is no lines with the current analytic account, the po is ne selected
The verification of the analytic account should only be done during the Detection phase of a candidate purchase line. At this moment we can check if the candidate line has the same analytic account and if so, select it or ignore it.

Video link to see the current behavior : https://youtu.be/Q4DXCCMzt0c

### DIFF
--- a/procurement_mto_analytic/models/__init__.py
+++ b/procurement_mto_analytic/models/__init__.py
@@ -3,3 +3,4 @@
 from . import sale_order_line
 from . import stock_move
 from . import stock_rule
+from . import purchase_order_line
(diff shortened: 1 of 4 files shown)


AaronHForgeFlow:
So you expect in the video that the third sales order to generate a third PO line in the PO that was already created, right? I agree that behavior makes more sense

adrienpeiffer:
Exactly, in this situation, it's not necessary to create a new purchase order because we already have an existing one for the same supplier.

AaronHForgeFlow (approved):
Code review LGTM

PR #455, opened 2022-05-17
[14.0][ADD] account_analytic_required_never

rousseldenis:
When creating account move lines, don't fill in analytic account (even if passed in parameters) if analytic policy is set to 'Never'


rousseldenis:
@sanrav @adrienpeiffer

adrienpeiffer (approved):
LGTM

rousseldenis:
@OCA/accounting-maintainers

rafaelbn:
I don't understand the difference with https://github.com/OCA/account-analytic/blob/14.0/account_analytic_required/models/account.py#L17

```
    property_analytic_policy = fields.Selection(
        selection=[
            ("optional", "Optional"),
            ("always", "Always"),
            ("posted", "Posted moves"),
            ("never", "Never"),
``` 

The question is:

- Without this module if account_analytic_required is set to "never" and the analytic account is set then de account.move has analytic
- With this module if account_analytic_required is set to "never" and the analytic account is set then de account.move has NOT analytic


Is it?

Thank you!

PR #462, opened 2022-06-07
[14.0] product_analytic_sale

alexis-via:
The readme gives an example of a scenario where this module product_analytic_sale is required.

### DIFF
--- a/product_analytic_sale/models/sale_order_line.py
+++ b/product_analytic_sale/models/sale_order_line.py
@@ -10,9 +10,12 @@ class SaleOrderLine(models.Model):
 
     def _prepare_invoice_line(self, **optional_values):
         vals = super()._prepare_invoice_line(**optional_values)
-        ana_account = self.product_id.product_tmpl_id._get_product_analytic_accounts()[
-            "income"
-        ]
-        if ana_account:
-            vals["analytic_account_id"] = ana_account.id
+        if self.product_id:
+            ana_account = (
+                self.product_id.product_tmpl_id._get_product_analytic_accounts()[
+                    "income"
+                ]
+            )
+            if ana_account:
+                vals["analytic_account_id"] = ana_account.id
         return vals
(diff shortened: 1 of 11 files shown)


clementmbr (approved):
LGTM

rvalyi (approved):

clementmbr:
@alexis-via I just found a bug in case the Sale Order Line as no product_id (type note). I added a commit to fix it.
cc @rvalyi

product_analytic_sale/models/sale_order_line.py
```diff
@@ -10,9 +10,12 @@ class SaleOrderLine(models.Model):
 
     def _prepare_invoice_line(self, **optional_values):
         vals = super()._prepare_invoice_line(**optional_values)
-        ana_account = self.product_id.product_tmpl_id._get_product_analytic_accounts()[
-            "income"
-        ]
-        if ana_account:
-            vals["analytic_account_id"] = ana_account.id
+        if self.product_id:
+            ana_account = (
```
rousseldenis, line 14: Nitpicking, but I'd prefer analytic_account than the short (less significant) term

rousseldenis:
It should be great to add a little test.

rafaelbn (approved):
This works 👍 

As @rousseldenis , it could be great a little test https://github.com/OCA/account-analytic/pull/462#issuecomment-1154109242

My curiosity / question is , why not improve product_analytic with this so we don't need glue modules? I complicated so we need another module

rousseldenis:
> This works +1
> 
> As @rousseldenis , it could be great a little test [#462 (comment)](https://github.com/OCA/account-analytic/pull/462#issuecomment-1154109242)
> 
> My curiosity / question is , why not improve product_analytic with this so we don't need glue modules? I complicated so we need another module

@rafaelbn This to avoid a strong dependency on sale in product_analytic

PR #466, opened 2022-06-21
[15.0][MIG] purchase_request_analytic

JuanyDForgeflow:
@ForgeFlow
* Assign an analytic account to the purchase request or to purchase request lines.


rousseldenis (approved):

AaronHForgeFlow (commented):
If I create several purchase request lines with an analytic account and then I change the analytic account in the purchase request I expect all the purchase request lines to change the analytic account to the new one. That is not happening.

PR #473, opened 2022-08-11
[13.0][IMP] purchase_analytic: adapt to newest framework.

LoisRForgeFlow:
Some of the "tricks" done in this module are no longer needed
and can be easily implemented with newest framework features:

* No need for an auxiliar `project_id2` field. User can set
  an analytic account with no lines and it is respected.
* Simplify onchange. Now update analytic line on the go (no
  need to save) which is a better UX because avoid unexpected
  changes on save.

Also re-label the field `project_id` to "Analytic Account" to
align with the typical label in newer versions of Odoo.
Tha label "Contract / Analytic" was last used in v8 (
https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L217).

@ForgeFlow

### DIFF
--- a/purchase_analytic/tests/test_purchase_analytic.py
+++ b/purchase_analytic/tests/test_purchase_analytic.py
@@ -19,7 +19,7 @@ class TestPurchaseAnalytic(TransactionCase):
     def test_analytic_account(self):
         """ Create a purchase order (create)
             Set analytic account on purchase
-            Check analytic account on line is set
+            Check analytic account and line is set
         """
         po = self.env["purchase.order"].create(
             {
@@ -48,7 +48,7 @@ class TestPurchaseAnalytic(TransactionCase):
     def test_project_id(self):
         """ Create a purchase order (new)
             Set analytic account on purchase
-            Check analytic account is on purchase
+            Check analytic account and line is set
         """
         po = self.env["purchase.order"].new(
             {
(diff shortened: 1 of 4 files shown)


MiquelRForgeFlow:
Since the test_project_id breaks, please avoid using new + onchange and use common.Form instead 😉

purchase_analytic/tests/test_purchase_analytic.py
```diff
@@ -19,7 +19,7 @@ def setUp(self):
     def test_analytic_account(self):
         """ Create a purchase order (create)
             Set analytic account on purchase
-            Check analytic account on line is set
+            Check analytic account ond line is set
```
MiquelRForgeFlow:
```suggestion
            Check analytic account one line is set
```

LoisRForgeFlow:
> Since the test_project_id breaks, please avoid using new + onchange and use common.Form instead wink

I was failing because there was a slight behavior change, now the analytic account in the lines is set without needing to save, which will lead to less confusion on users.

I just updated the test.

purchase_analytic/tests/test_purchase_analytic.py
LoisRForgeFlow: it is "**a**nd", fixed :)

purchase_analytic/tests/test_purchase_analytic.py
```diff
@@ -48,7 +48,7 @@ def test_analytic_account(self):
     def test_project_id(self):
         """ Create a purchase order (new)
             Set analytic account on purchase
-            Check analytic account is on purchase
+            Check analytic account ond line is set
```
MiquelRForgeFlow: here too
```suggestion
            Check analytic account and line is set
```

LoisRForgeFlow: corrected!

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
Functional tests. Same behavior, better performance.

PR #474, opened 2022-08-16
[15.0][ADD] migrate sale_stock_analytic module from v14 to v15

Urvisha-OSI:
https://pm.opensourceintegrators.com/web#id=24145&action=324&model=helpdesk.ticket&view_type=form&menu_id=218


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #478, opened 2022-08-22
[15.0][IMP] purchase_analytic: adapt to newest framework.

LoisRForgeFlow:
Forward port of #473 

Some of the "tricks" done in this module are no longer needed
and can be easily implemented with newest framework features:

* No need for an auxiliar `project_id2` field. User can set
  an analytic account with no lines and it is respected.
* Simplify onchange. Now update analytic line on the go (no
  need to save) which is a better UX because avoid unexpected
  changes on save.

Also re-label the field `project_id` to "Analytic Account" to
align with the typical label in newer versions of Odoo.
Tha label "Contract / Analytic" was last used in v8 (
https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L217).

@ForgeFlow


AaronHForgeFlow (approved):
Thank you for the forward port!

PR #482, opened 2022-09-23
[14.0][FIX] account_analytic_sequence: Make sequence multi-company by default

pedrobaeza:
For avoiding an error in secondary companies when creating analytic accounts there.

Steps to reproduce:

- Have an instance with at least 2 companies.
- Install module with one company as main.
- Switch to the other company.
- Create an analytic account.

It will give an error.

No regression test is provided this time as it will require a lot of code for something fixed just with not filling a field.

@Tecnativa


victoralmau (approved):

adrienpeiffer (approved):

pedrobaeza:
Thanks! Here is the fw-port to v15: #483

PR #484, opened 2022-09-28
[14.0][ADD] purchase_analytic_required

rousseldenis:
Replaces #449


rousseldenis:
@sanrav @OCA/accounting-maintainers

purchase_analytic_required/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Denis Roussel <denis.roussel@acsoe.eu>
```
adrienpeiffer: Your new company ?

adrienpeiffer (approved):
Just one typo to fix.

xavier-bouquiaux (approved):
LGTM

acsonefho (approved):
Code LGTM

acsonefho (changes requested):
I think you should put the `required=True` on the field.

rousseldenis:
@sbidoul As discussed, it's maybe better to do a constrains() and an ir.config.parameter to activate the behaviour.

sbidoul:
I generally think that modules that change a standard behaviour should have a config settings to activate the change.
Otherwise we have all sorts of issues in tests with incompatible modules, etc.

Adding a required field without default value is such a change. Making the change in the view keeps compatibility - now we may not need an OCA module just to make a field required in a view.

Related thought: we may want to make the analytic account required only when confirming the PO? (I don't know your use cases though, so let's not overengineer things).

acsonefho:
> I generally think that modules that change a standard behaviour should have a config settings to activate the change. Otherwise we have all sorts of issues in tests with incompatible modules, etc.

I agree with you but in this case, it's the name of the module and it does nothing else.
IMO doing all this adaptation only because we could have issue in tests is not a good idea. We can execute it in a specific job.
Adding a module to enable his only feature via a configuration seems very strange.

PR #441, opened 2022-03-07
[14.0][ADD] stock_picking_type_analytic

rousseldenis:
Depends on:

- [ ] #406

PR #406, opened 2021-10-20
[14.0][ADD] stock_picking_analytic

rousseldenis:
Depends on:

- [x] https://github.com/OCA/account-analytic/pull/423

PR #486, opened 2022-09-29
[14.0][ADD] stock_picking_type_analytic

rousseldenis:
Replaces #441 

Depends on:

- [ ] #406


stock_picking_type_analytic/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+        Get analytic account from picking type if set
+        """
+        defaults = self.default_get(["name", "picking_type_id"])
+        picking_type = self.env["stock.picking.type"].browse(
+            vals.get("picking_type_id", defaults.get("picking_type_id"))
+        )
+        if "analytic_account_id" not in vals and picking_type.analytic_account_id:
+            vals.update(
+                {
+                    "original_analytic_account_id": picking_type.analytic_account_id.id,
```
apineux, line 23: Shouldn't it be analytic_account_id instead of original_analytic_account_id?

rousseldenis: https://github.com/OCA/account-analytic/blob/14.0/stock_picking_analytic/models/stock_picking.py#L11

rousseldenis: But indeed, this highlight the fact tests don't pass on that line.

apineux (approved):

PR #493, opened 2022-10-18
[13.0][IMP] product_analytic: move if condition to separate function

JaumeBforgeFlow:
Moving the condition for the create method on account.move.line on module product_analytic to its own method.
This will allow to change/extend this condition without needing to inherit the full create function and allowing to keep the logic for using the product analytic account.
Also it will only write the analytic account from the product if there is actualy one.

### DIFF
--- a/account_analytic_sequence/data/sequence.xml
+++ b/account_analytic_sequence/data/sequence.xml
@@ -7,6 +7,6 @@
         <field name="code">account.analytic.account.code</field>
         <field name="prefix">AA</field>
         <field name="padding">3</field>
-        <field name="company_id" eval="False"/>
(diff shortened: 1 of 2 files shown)


JordiMForgeFlow (approved):
LGTM 👍🏼 Hooks are always nice!

jguenat:
Hello,

I did some work there #512 to add analytic tags support. I also added hooks.
If it gets merged I could backport it to 13 and 14  and we could close this one?

JordiMForgeFlow:
Hi @jguenat 

I think it makes sense, that way we can also benefit from your other improvements. Thank you! :)

PR #494, opened 2022-10-29
16.0 mig mrp analytic

KM-joel:
Migration to version 16.0

- deletion of the inherit of the mrp production object because there is already an analytic_account_id field
- deletion of the inherit form and list view of the mrp.production object in order to place an analytic_account_id field that already exists in the odoo base 
- I also removed the analytics account id filter to place in group by because there is already a grouping filter by this field
- add default plan and company when creating an analytical account in the test


auroraeosrose:
There's a bug in stock_move.py

_generate_valuation_lines_data in base odoo in 16 has an additional parameter svl_id that's missing

```
rslt = super(StockMove, self)._generate_valuation_lines_data(partner_id, qty, debit_value, credit_value, debit_account_id, credit_account_id, svl_id, description)
TypeError: StockMove._generate_valuation_lines_data() takes 8 positional arguments but 9 were given

```

dreispt:
Superseded by #553

AaronHForgeFlow:
Closed as superseded

PR #497, opened 2022-11-16
[14.0][ADD] module account_move_update_analytic

remi-filament:
Add new module allowing to update analytic account and tags on posted moves
(Note that although model has changed, this is now allowed in v16)

Moved to this repo from https://github.com/OCA/account-financial-tools/pull/1485

@rafaelbn  @fclementic2c I have mode module here as suggested, also I have added the handling of analytic tags.

Let me know your thoughts !


rafaelbn (commented):
Hello @remi-filament 

Tested, it works. Maybe I should test it deeper 😄 

Doubts about module name:
`account_move_update_analytic` vs `account_move_analytic_update`

And finally a re-flexion trying to realize the best UX for accountants and financial responsible that I have recorded 😃 

https://www.loom.com/share/a8c9edbb3bac493084f871e2df33a6a6

What do you think @remi-filament  @fclementic2c ?

I'm just asking to know your opinion and thougts in order to get the best UX for the module.

We can of course if we agree contribute not only reviewing.

After that obviusly I will approve and merge!

Thank you @remi-filament for this great module and contribution!!!! ❤️ 

cc @moduon @Shide (please review tech+UX) MT-1606

account_move_update_analytic/views/account_move_line_view.xml
```diff
@@ -0,0 +1,23 @@
...
+<odoo>
+
+    <record model="ir.ui.view" id="account_move_line_tree_view">
+        <field name="name">account.move.line.tree</field>
+        <field name="model">account.move.line</field>
+        <field name="inherit_id" ref="account.view_move_line_tree" />
+        <field name="arch" type="xml">
+            <field name="analytic_account_id" position="after">
+                <button
+                    name="account_move_update_analytic.action_view_account_move_update_analytic"
```
Shide, line 13: Can be simplified to
```suggestion
                    name="action_view_account_move_update_analytic"
```

account_move_update_analytic/views/account_move_line_view.xml
```diff
@@ -0,0 +1,23 @@
...
+                    string="Update Analytic"
+                    type="action"
+                    groups="analytic.group_analytic_accounting,analytic.group_analytic_tags"
+                    attrs="{'invisible': [('parent_state', '!=', 'posted'),]}"
```
Shide, line 34: Last comma of domain is unnecesary. _It's only a suggestion_
```suggestion
                    attrs="{'invisible': [('parent_state', '!=', 'posted')]}"
```

account_move_update_analytic/views/account_move_view.xml
```diff
@@ -0,0 +1,38 @@
...
+            <xpath
+                expr="//field[@name='invoice_line_ids']/tree/field[@name='analytic_tag_ids']"
+                position="after"
+            >
+                <button
+                    name="account_move_update_analytic.action_view_account_move_update_analytic"
+                    string="Update Analytic"
+                    type="action"
+                    groups="analytic.group_analytic_accounting,analytic.group_analytic_tags"
+                    attrs="{'invisible': [('parent.state', '!=', 'posted'),]}"
```
Shide: Use `parent_state` instead of `parent.state`. It's already computed stored=True and the final comma. You do this before
```suggestion
                    attrs="{'invisible': [('parent_state', '!=', 'posted')]}"
```

account_move_update_analytic/views/account_move_view.xml
```diff
@@ -0,0 +1,38 @@
...
+            <xpath
+                expr="//field[@name='line_ids']/tree/field[@name='analytic_tag_ids']"
+                position="after"
+            >
+                <button
+                    name="account_move_update_analytic.action_view_account_move_update_analytic"
+                    string="Update Analytic"
+                    type="action"
+                    groups="analytic.group_analytic_accounting,analytic.group_analytic_tags"
+                    attrs="{'invisible': [('parent.state', '!=', 'posted'), ('display_type', 'in', ('line_section', 'line_note'))]}"
```
Shide, line 34: Use `parent_state` instead of `parent.state`. It's already computed stored=True. You do this before
```suggestion
                    attrs="{'invisible': [('parent_state', '!=', 'posted'), ('display_type', 'in', ('line_section', 'line_note'))]}"
```

account_move_update_analytic/wizards/account_move_update_analytic.py
```diff
@@ -0,0 +1,55 @@
...
+    new_analytic_account_id = fields.Many2one(
+        "account.analytic.account", string="New Analytic Account", check_company=True
+    )
+    new_analytic_tag_ids = fields.Many2many(
+        "account.analytic.tag", string="New Analytic Tags", check_company=True
+    )
+
+    @api.model
+    def default_get(self, fields):
+        rec = super(AccountMoveUpdateAnalytic, self).default_get(fields)
```
Shide: You can use `super()` as it is. _It's only a suggestion_

account_move_update_analytic/wizards/account_move_update_analytic.py
```diff
@@ -0,0 +1,55 @@
...
+        context = dict(self._context or {})
+        active_id = context.get("active_id", False)
```
Shide: You can use `self.env.context` and remove the line above (`context = dict(self._context or {})`)
```suggestion
        active_id = self.env.context.get("active_id")
```

account_move_update_analytic/wizards/account_move_update_analytic.py
```diff
@@ -0,0 +1,55 @@
...
+                "line_id": active_id,
+                "current_analytic_account_id": aml.analytic_account_id.id,
+                "current_analytic_tag_ids": [(6, 0, aml.analytic_tag_ids.ids or [])],
+                "company_id": aml.company_id.id,
+            }
+        )
+        return rec
+
+    def update_analytic_lines(self):
+        self.line_id.analytic_line_ids.unlink()
```
Shide, line 43: `self.ensure_one()` to ensure only one wizard is being called at one time
```suggestion
        self.ensure_one()
        self.line_id.analytic_line_ids.unlink()
```

account_move_update_analytic/wizards/account_move_update_analytic.py
```diff
@@ -0,0 +1,55 @@
...
+        if self.user_has_groups("analytic.group_analytic_accounting"):
+            self.line_id.write(
```
Shide: Can be simplified to
```suggestion
            self.line_id.analytic_account_id = self.new_analytic_account_id
```

account_move_update_analytic/wizards/account_move_update_analytic.py
```diff
@@ -0,0 +1,55 @@
...
+                {"analytic_account_id": self.new_analytic_account_id.id or False}
+            )
+        if self.user_has_groups("analytic.group_analytic_tags"):
+            self.line_id.write(
```
Shide, line 47: Same as above and not need last `[]` because .ids always return `[]` if it's an empty recordset.
I think it will work without .ids
I'm not 100% sure if the `[(6, 0, ...)]` it's needed in this case, but trying it will be great.
```suggestion
            self.line_id.analytic_tag_ids = self.new_analytic_tag_ids
```

account_move_update_analytic/wizards/account_move_update_analytic_view.xml
```diff
@@ -0,0 +1,63 @@
...
+                    groups="analytic.group_analytic_accounting"
+                >
+                    <group>
+                        <field name="current_analytic_account_id" readonly="1" />
+                    </group>
+                    <group>
+                        <field name="company_id" invisible="1" />
+                        <field
+                            name="new_analytic_account_id"
+                            domain="['|', ('company_id', '=', False), ('company_id', '=', company_id)]"
```
Shide: `check_company=True` is present in `new_analytic_account_id` field. It's really needed?

account_move_update_analytic/wizards/account_move_update_analytic_view.xml
```diff
@@ -0,0 +1,63 @@
...
+                            name="current_analytic_tag_ids"
+                            readonly="1"
+                            widget="many2many_tags"
+                        />
+                    </group>
+                    <group>
+                        <field name="company_id" invisible="1" />
+                        <field
+                            name="new_analytic_tag_ids"
+                            domain="['|', ('company_id', '=', False), ('company_id', '=', company_id)]"
```
Shide: Same as above

Shide (changes requested):
**Super really great work!**

Some things can be tweaked but are minor fixes.
The most important for me is the use of `parent_state` instead of `parent.state` and adding the button in the `account.view_move_line_tree_grouped` view

_This is a code review without using the module on runboat_

account_move_update_analytic/views/account_move_line_view.xml
```diff
@@ -0,0 +1,23 @@
...
+                />
+            </field>
+        </field>
+    </record>
+
```
Shide, line 39: Also add the button in the tree grouped view for allow Accountants centralize their work

```suggestion
    <record model="ir.ui.view" id="view_move_line_tree_grouped">
        <field name="name">account.move.line.tree.grouped</field>
        <field name="model">account.move.line</field>
        <field name="inherit_id" ref="account.view_move_line_tree_grouped" />
        <field name="arch" type="xml">
            <field name="analytic_account_id" position="after">
                <button
                    name="account_move_update_analytic.action_view_account_move_update_analytic"
                    string="Update Analytic"
                    type="action"
                    groups="analytic.group_analytic_accounting,analytic.group_analytic_tags"
                    attrs="{'invisible': [('parent_state', '!=', 'posted'),]}"
                />
            </field>
        </field>
    </record>
```

Shide (changes requested):
The missing view in `view_move_line_tree_grouped`

remi-filament:
Hi @rafaelbn sorry for the delay in coming back to you !
Very busy weeks over here, thank you for taking time to review it and even record a video !
I get your point about having the updates made by accountant in general ledger or accounting menus, however I have a different view because :
1. With French regulations, it is not allowed to edit an invoice once it has been posted (you can only refund and create a new one), therefore in Odoo when configuring FR company, you cannot edit account_move_lines once posted, nor can you bring an invoice back to draft
2. My customers are not accountant, but rather want to go back to their invoice to correct analytic account when they get discrepancy analyzing margins on projects

I guess we can still add button on account_move_line tree views.
Also I did not manage to make the column hideable but I should look again, maybe it is only because I forgot to give a name to column...

Thanks !

account_move_update_analytic/wizards/account_move_update_analytic_view.xml
remi-filament: I am not sure, I should try without, but this is how it is done on Odoo code here : https://github.com/odoo/odoo/blob/a99055180c6c8c87603d4beca2156c9e6898ffa4/addons/account/views/account_move_views.xml#L756-L764

account_move_update_analytic/views/account_move_view.xml
remi-filament: I then needed to add this parent_state field as invisible before in view, but done

account_move_update_analytic/wizards/account_move_update_analytic.py
remi-filament: I removed the or [] but had to keep the [(6,0,xx)] otherwise it does not empty the field if you remove analytic tags

remi-filament:
Thanks a lot @Shide for your comments, I updated accordingly !
I made a separate commit to that you can see the additions, however I can squash these into one if OK for you

@rafaelbn as proposed by @Shide I added the button in view_move_line_tree_grouped as well which was if I understood correctly what you were asking for !

Also I did not manage to use show/hide to hide the column with the button, nor adding a title to that column, I am not sure it is even feasible ?

Let me know
Best Regards

rafaelbn:
Thank you @remi-filament !

I found a BUG, check (2 min): https://www.loom.com/share/a274be3a321a4ae3986ee705cf7e94ea

rafaelbn:
> With French regulations, it is not allowed to edit an invoice once it has been posted (you can only refund and create a new one), therefore in Odoo when configuring FR company, you cannot edit account_move_lines once posted, nor can you bring an invoice back to draft

@remi-filament , Spain is similar, I mean, and invoice is not accounting even Odoo made same model. Change an account is not change an invoice, invoicing means more amount and taxes, accounting go in other hand. Futhermore, the change is not allowed for full invoice just for income or outcome (expenses).

Accounting Reclassification si common.

For example, when we invoice and send it in real time to administration you are not sending accounts, just amount base, tax, vat, types... the accounting is later, usually once a year.

I'm working for multi-nationals companies with monthly closing accounting, they are really strict, but accountants sometimes fails and they have to reclassify accounting before closing (and should be really fast), and this is legal if you don't change taxes or amounts that you previously have sent.

Do you see the differences?

rafaelbn (changes requested):
Please check this. I found a BUG, check (2 min): https://www.loom.com/share/a274be3a321a4ae3986ee705cf7e94ea

remi-filament:
Hi @rafaelbn 
Thanks for reporting the issue, of course we should not break anything.
I can reproduce on runboat but not on my local instance with only this module installed from account-analytic OCA repo. 
Also, I am not updating account_move_line model code, so I would not understand how my module could imply the bug you see.
So I suppose the issue comes from some other module from that repo...

I went quickly over the code of all modules in this repo but cannot find the culprit...

Also, regarding your other point, I am not used to update these fields as an accountant but I guess the addition of the button on account move line grouped tree view is what you expected ?
FYI, in France with hash enabled on customer invoices, you cannot update even account once the invoice is posted (you get the following error : You cannot edit the following fields due to restrict mode being activated on the journal: debit, credit, account_id, partner_id.), which is why I never was allowed to do it !

remi-filament:
I managed to find the culprit I think, when uninstalling account_analytic_tag_default the expected behavior is back.
Probably because of the compute of account_analytic_tag_ids here : https://github.com/OCA/account-analytic/blob/ce727f15024386730988f6180b3ead4d73efd0ea/account_analytic_tag_default/models/account_move.py#L18

rafaelbn:
Hello @remi-filament ,

> Also, regarding your other point, I am not used to update these fields as an accountant but I guess the addition of the button on account move line grouped tree view is what you expected ?

No by the moment.

On monday we will return to this PR to get it reviewed and merge next week 👍 

Thank you @remi-filament !! 🙏

Shide:
> Thanks a lot @Shide for your comments, I updated accordingly ! I made a separate commit to that you can see the additions, however I can squash these into one if OK for you
> 
> @rafaelbn as proposed by @Shide I added the button in view_move_line_tree_grouped as well which was if I understood correctly what you were asking for !
> 
> Also I did not manage to use show/hide to hide the column with the button, nor adding a title to that column, I am not sure it is even feasible ?
> 
> Let me know Best Regards

I think it's not possible right now.

Shide:
@remi-filament I have one question:

Why not make the field not-readonly and allow the user edit directly from the tree lines? 
Is there any limitation that forces us to show a wizard? Because the wizard only allows to modify one line at a time.

The multi selection + inline tree editing allows to edit all the lines with the same analytic accounts at once.

remi-filament:
Hi @Shide thanks for looking again into it !


> Why not make the field not-readonly and allow the user edit directly from the tree lines?
> Is there any limitation that forces us to show a wizard? Because the wizard only allows to modify one line at a time.

Actually, the problem is not only updating analytic account or tags on account_move_lines, this could indeed be done as you suggest, however you need to update / delete / create account_analytic_lines and/or account_analytic_tags which are related to these account_move_lines.
We could still have a function with a depends that recreate lines if account_move_line is updated but I think it is better to get a wizard (+ you also have a second verification with the wizard, since you need to click on 2 buttons instead of only editing and Save).

I agree though that we could go further and make a mass edit wizard that would update analytic account and analytic tags on multiple lines at once, but this was not my initial use case, and I like to know that I am changing from this account to this new account which I think is more difficult to get if you change multiple lines at once, so that would need more thinking about UX !

Shide:
> Hi @Shide thanks for looking again into it !
> 
> > Why not make the field not-readonly and allow the user edit directly from the tree lines?
> > Is there any limitation that forces us to show a wizard? Because the wizard only allows to modify one line at a time.
> 
> Actually, the problem is not only updating analytic account or tags on account_move_lines, this could indeed be done as you suggest, however you need to update / delete / create account_analytic_lines and/or account_analytic_tags which are related to these account_move_lines. We could still have a function with a depends that recreate lines if account_move_line is updated but I think it is better to get a wizard (+ you also have a second verification with the wizard, since you need to click on 2 buttons instead of only editing and Save).
> 
> I agree though that we could go further and make a mass edit wizard that would update analytic account and analytic tags on multiple lines at once, but this was not my initial use case, and I like to know that I am changing from this account to this new account which I think is more difficult to get if you change multiple lines at once, so that would need more thinking about UX !

Ok, then It's fine for me 😄

Shide (approved):
LGTM

rafaelbn (approved):
Thanks!🙂 👍 

@fclementic2c we are going to migrate to v15 if you would like to test it for any improvement, you are really welcome

/ocabot merge nobump

PR #507, opened 2022-12-14
[FIX] [14.0] Create analytic lines if is posted

Shide:
Only create analytic lines if move_id is posted.
Track analytic account and tags fields

@rafaelbn @remi-filament please review

MT-1778 @moduon

### DIFF
--- a/account_move_update_analytic/__init__.py
+++ b/account_move_update_analytic/__init__.py
@@ -1,1 +1,2 @@
+from . import models
 from . import wizards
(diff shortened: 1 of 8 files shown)


rafaelbn (approved):
👍

remi-filament (approved):
Thanks @Shide, makes sense to add another verification here (the only check was on the view where button was shown only for posted entries).
Maybe you can also use parent_state instead of move_id.state ?

Shide:
I've added the tracking fields suggestion https://github.com/OCA/account-analytic/pull/508#pullrequestreview-1218026006

rafaelbn (approved):
/ocabot merge patch

PR #510, opened 2022-12-16
[15.0][FIX][IMP] purchase_analytic: Fix deletion of analytic_id by onchange method

bosd:
The analytic accounts are corretly set with the compute method.
When using the GUI, the analytic account on the orderline was overriden by the onchange method.

This PR adds the analytic account to the onchange method.


bosd:
Please remove stale again :)

PR #521, opened 2022-12-29
[15.0][MIG]stock_quant_analytic: Migration to 15.0

bizzappdev:


AaronHForgeFlow:
The stock_inventory module is going back to live here: https://github.com/OCA/stock-logistics-warehouse/pull/1512

I am not sure, but maybe should this module depend on that one? or in other case be renamed to stock_quant_analytic to prevent confusion?

bizzappdev:
@AaronHForgeFlow  indeed module name should be changed to stock_quant or we have to introduce the stock_inventory dependency. I am not the proper person to decide the same. might be maintainer can decide. 
Unfortunately, this module does not have a maintainer set. so might be @rousseldenis can help here?

rousseldenis:
> @AaronHForgeFlow indeed module name should be changed to stock_quant or we have to introduce the stock_inventory dependency. I am not the proper person to decide the same. might be maintainer can decide. Unfortunately, this module does not have a maintainer set. so might be @rousseldenis can help here?

I would say that the name should be kept only if its aim is to put analytic dimensions to stock quant only in inventories.

If analytics are put in every flow, so, the name should maybe changed to stock_quant_analytic yes.

bizzappdev:
ready for the review

PR #526, opened 2023-01-19
[14.0][FIX] mrp_analytic_sale_project: found case not carrying Analytic as expected

dreispt:
See https://github.com/odoo/odoo/blob/14.0/addons/sale_mrp/models/sale.py


patrickrwilson (approved):
LGTM

patrickrwilson:
Also added a quick fix for when no MO's are created

PR #527, opened 2023-01-21
[16.0][MIG] stock_picking_analytic: Migration to 16.0

yostashiro:
* Switch field from `analytic_account_id` to `analytic_distribution`.
* ~~Change depends referenced field from `move_ids_without_package` to `move_ids`.  `move_ids_without_package` is a non-stored compute field and therefore is not appropriate to be used as a trigger (you get a warning message in running tests without this change).~~ Add dummy search arg to `move_ids_without_package` to bypass warning.
* Remove `groups="analytic.group_analytic_accounting"` from the `analytic_distribution` field added in the picking form, because it would cause an error `Field 'analytic_distribution' used in context ({'default_company_id': company_id, 'default_date': scheduled_date, 'default_date_deadline': date_deadline, 'picking_type_code': picking_type_code, 'default_picking_id': id, 'form_view_ref': 'stock.view_move_form', 'address_in_id': partner_id, 'default_picking_type_id': picking_type_id, 'default_location_id': location_id, 'default_location_dest_id': location_dest_id, 'default_partner_id': partner_id, 'default_analytic_distribution': analytic_distribution}) is restricted to the group(s) analytic.group_analytic_accounting` in loading the view.  Looks like this refactoring https://github.com/odoo/odoo/commit/0501bbd62e517f6c215d9e7e36d61747c7f5816b has added the check.  Not sure there is a good way to still add the group to the field, however there is not much point doint it since the module wouldn't be installed without enabling the analytic accounting.  -> Add an invisible `analytic_distribution` field without group assignment (thanks @luisg123v ).
* ~~Extension of the search view to add `analytic_distribution` field is disabled as the search on json fields do not work in a meaningful manner as of now.~~

@qrtl


rousseldenis:
@yostashiro Shouldn't this deserves a migration script ?

yostashiro:
> @yostashiro Shouldn't this deserves a migration script ?

@rousseldenis I was thinking no, as I thought the whole point of having analytic account (analytic distribution now) in the picking was to propagate the value to stock moves, and few people would care about the data persistency of itself (but I might be wrong).  If anything, migration script should be added in https://github.com/OCA/account-analytic/tree/16.0/stock_analytic for analytic distribution of stock moves?

desdelinux (approved):
Functional review is ok.

cc @rolandojduartem @rousseldenis @luisg123v

desdelinux:
> > @yostashiro Shouldn't this deserves a migration script ?
> 
> @rousseldenis I was thinking no, as I thought the whole point of having analytic account (analytic distribution now) in the picking was to propagate the value to stock moves, and few people would care about the data persistency of itself (but I might be wrong). If anything, migration script should be added in https://github.com/OCA/account-analytic/tree/16.0/stock_analytic for analytic distribution of stock moves?

@rousseldenis I agree with this

rolandojduartem:
LGTM +1

rolandojduartem (approved):

stock_picking_analytic/models/stock_picking.py
```diff
@@ -0,0 +1,57 @@
...
+        inverse="_inverse_analytic_distribution",
+        states={
+            "done": [("readonly", True)],
+            "cancel": [("readonly", True)],
+        },
+    )
+    # Assign a dummy search arg to the field to avoid triggering the warning.
+    # Without this, use of move_ids_without_package in @api.depends() of the compute
+    # method would trigger a warning from
+    # https://github.com/odoo/odoo/blob/ae94f14a844352f66490ec326fe2d1a716023891/odoo/fields.py#L808-L812  # noqa
```
luisg123v: If you use a shorter hash, line won't be as long and noqa won't be required, e.g.:

```suggestion
    # https://github.com/odoo/odoo/blob/ae94f14a/odoo/fields.py#L808-L812
```

stock_picking_analytic/tests/test_picking_analytic_account.py
```diff
@@ -0,0 +1,120 @@
...
+            {
+                "analytic_distribution": self.analytic_distribution,
+            }
+        )
+        self.picking.write(
+            {
+                "move_ids": [
+                    (
+                        0,
+                        0,
```
luisg123v: Prefer the Command syntax

Same below

stock_picking_analytic/tests/test_picking_analytic_account.py
```diff
@@ -0,0 +1,120 @@
...
+            picking.move_ids[0].analytic_distribution,
+            self.analytic_distribution,
+        )
+
+    def test_compute_analytic_distribution(self):
+        """Set analytic distribution on move
+        Check analytic distribution on picking is set
+        """
+        picking = self.picking
+        self.assertNotEqual(len(picking.move_ids), 0)
```
luisg123v:
```suggestion
        self.assertTrue(picking.move_ids)
```

Same below

stock_picking_analytic/views/stock_picking.xml
```diff
@@ -0,0 +1,64 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <record model="ir.ui.view" id="stock_picking_form_view">
+        <field name="name">stock.picking.form (in stock_picking_analytic)</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group/group/field[@name='picking_type_id']" position="after">
+                <!-- groups="analytic.group_analytic_accounting" has been removed from
```
luisg123v: You just need to add the field again as invisible, so the field is included no matter if current user belongs to the group

luisg123v (changes requested):
You're including commits that are not merged in the stable branch, some of which are even WIP:
- 452e4260144a3a6f510a6004f6ce86b2e7be156e
- 74e5d890f0498717a14fdb54443a5a18eaf8b83e

Those are migration commits to 15.0, but this module is not migrated to 15.0 yet. I'd suggest either:
- removing those ones from this migration
- Or  creating (first) a migration PR to 15.0 and, after that one is merged, resume this one

On the other hand:

"pre-commit" -> "pre-commit stuff"

luisg123v:
PR description is a comprehensive one, it would be a shame if it's not included also in commit description.

desdelinux:
@yostashiro Hello, you can apply the suggested changes

yostashiro:
@luisg123v @desdelinux Thanks for your review comments. Will try and follow up on them but give me some time. A bit swamped at the moment.

rousseldenis:
@yostashiro Any news on this ?

yostashiro:
I plan to update the PR this weekend.

yostashiro:
@luisg123v @desdelinux @rousseldenis Thanks for waiting. Just updated the PR. I believe I took care of all the points in the feedback.

desdelinux (approved):
Functional Review is ok.

cc @luisg123v @rousseldenis

rolandojduartem:
@luisg123v could you review this PR?

stock_picking_analytic/views/stock_picking.xml
```diff
@@ -0,0 +1,73 @@
...
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group/group/field[@name='picking_type_id']" position="after">
+                <field
+                    name="analytic_distribution"
+                    widget="analytic_distribution"
+                    groups="analytic.group_analytic_accounting"
+                />
+                <!-- Just to prevent error in subsequent context update - assigining
```
luisg123v: Typo:  assigining -> assigning

But I think this comment should be removed, this is not a special case but just the way v16 works. Most of views containing fields with groups (e.g. almost all views with the `company_id` field) define the same field as invisible.

A similar case occurred in previous versions when using a field in attrs that was not already present in the view

stock_picking_analytic/views/stock_picking.xml
```diff
@@ -0,0 +1,73 @@
...
+                groups to the field would trigger an error in 16.0, therefore we add
+                an invisible field that is accessible to everyone. -->
+                <field
+                    name="analytic_distribution"
+                    widget="analytic_distribution"
```
luisg123v: Since it's invisible, widget is meaningless

stock_picking_analytic/tests/test_picking_analytic_account.py
```diff
@@ -21,82 +26,81 @@ def setUpClass(cls):
...
-        self.assertNotEqual(len(picking.move_ids_without_package), 0)
+        self.assertTrue(picking.move_ids_without_package)
         self.assertNotEqual(
-            picking.move_ids_without_package[0].analytic_account_id,
-            self.analytic_account,
+            picking.move_ids_without_package[0].analytic_distribution,
+            self.analytic_distribution,
         )
-        picking.move_ids_without_package.write(
+        picking.move_ids_without_package.update(
```
luisg123v: Why update instead of write?

write is preferred, because update writes fields one by one

Same below

luisg123v (changes requested):

stock_picking_analytic/tests/test_picking_analytic_account.py
yostashiro: `write()` would throw an error for `picking.move_ids_without_package` not having the database record when tests started.  I have switched to assign `move_ids` instead of `move_ids_without_package` to the picking in the setup to use `write()` in tests.

desdelinux (approved):
Functional Review is ok.

@rousseldenis

desdelinux:
@OCA/accounting-maintainers This one is ready

desdelinux (approved):

rolandojduartem (approved):

luisg123v (approved):
LGTM :+1:

luisg123v:
@moylop260,

Could you review/merge, please?

Regards,

PR #506, opened 2022-12-06
[MIG] stock_analytic: Migration to 16.0

nicomacr:

PR #528, opened 2023-01-23
[16.0][IMP] stock_analytic: add validation on analytic distribution

yostashiro:
To follow up on https://github.com/OCA/account-analytic/pull/506#issuecomment-1375161004.

- Add validation on analytic distribution according to how it is done in other models (account.move.line, sale.order.line, etc.).
- Update readme to reflect the current design.
- Other cosmetic adjustments.

@qrtl

### DIFF
--- a/stock_analytic/README.rst
+++ b/stock_analytic/README.rst
@@ -25,7 +25,7 @@ Stock Analytic
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-Adds an analytic account and analytic tags in stock move to be able to get
+Adds Analytic Distribution field in stock move to be able to get
 analytic information when generating the journal items.
 
 **Table of contents**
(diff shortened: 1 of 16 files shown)


kanda999 (approved):
Function review:
Analystic description validated in stock.picking

AungKoKoLin1997 (approved):
Function review:
Tested in runboat. It worked as expected.

AaronHForgeFlow (approved):
Code review LGTM. Thanks for the fix.

AaronHForgeFlow:
Just one last comment before the merge. I tried to validate a picking for a product with automated inventory valuation and with a stock move with the analytic distribution set I get the validation error "One or more lines require a 100% analytic distribution." I am not sure if it is an actual issue or a misconfiguration in the analytic plan or it is an actual issue. Does that make sense? here is the link to the picking in runboat: http://oca-account-analytic-16-0-pr528-06cc4c3c2c31.runboat.odoo-community.org/web?debug=1#id=39&cids=1&menu_id=168&action=319&active_id=1&model=stock.picking&view_type=form

yostashiro:
@AaronHForgeFlow I suppose that was coming from the mandatory default applicability set on a plan.
image

I guess this would try to do a validation on the generated account move lines (not on the stock moves where we make incoming moves out of validation scope), and therefore the validation error.  I just changed the default applicability of the plan to optional and I was able to validate the incoming picking.

AaronHForgeFlow (approved):
Functional review

AaronHForgeFlow:
I think 3 weeks is enough time for others to give their opinions and this is about to be merged now.

PR #532, opened 2023-01-29
[15.0][IMP] account_move_update_analytic: hide button column when not posted

ghost:
@remi-filament @Shide 
I changed the view slightly to hide the button column when the account move is not posted.

### DIFF
--- a/account_move_update_analytic/views/account_move_view.xml
+++ b/account_move_update_analytic/views/account_move_view.xml
@@ -19,7 +19,7 @@
                     string="Update Analytic"
                     type="action"
                     groups="analytic.group_analytic_accounting,analytic.group_analytic_tags"
-                    attrs="{'invisible': [('parent_state', '!=', 'posted')]}"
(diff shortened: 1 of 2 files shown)


ghost:
What do you think about setting the default new values = current values?
Then the user can change what should change, and the rest will be the same as before.
I did this in a new commit now.

Shide (approved):
LGTM, just fix the pre-commit issue

remi-filament (approved):
Thanks, looks better now

rafaelbn:
Please @appstogrow , fix the pre-commit issue 😄 thank you! after that we can merge!

norlinhenrik:
@rafaelbn I don't understand the pre-commit error message. Could you please have a look? Do you think it is related to this PR?

Shide:
It's ok for me. The pre-commit issue cannot be fixed by us and the code is well pre-commited.
@rafaelbn if you can merge... :)
LGTM

rafaelbn (approved):
/ocabot merge minor

PR #536, opened 2023-02-11
[16.0][ADD] mrp_stock_analytic

yostashiro:
Allows to assign the analytic distribution in the manufacturing order, which should
propagate the value to the corresponding field of the related component stock moves.

This function can be useful when cost analysis needs to be done on flushed components.

This module depends on OCA module stock_analytic.

I am aware that there is mrp_analytic module in this repo and there is a 16.0 migration PR https://github.com/OCA/account-analytic/pull/494, however I think this module kind of have lost the reason to exist in 15.0 when Odoo added `analytic_account_id` to mrp.production, and now it seems the functions around this field are unsupported (ref. https://github.com/odoo/odoo/pull/109910). Therefore this proposal with another name, based on the analytic distribution concept.

This PR depends on https://github.com/OCA/account-analytic/pull/528.

@qrtl


rinaldifirdaus (approved):
Functional test: working as expected

AaronHForgeFlow:
LG! Maybe the features here can be included in the mrp_analytic module, instead of doing a new module, WDYT?

yostashiro:
> LG! Maybe the features here can be included in the mrp_analytic module, instead of doing a new module, WDYT?

@AaronHForgeFlow That was my initial thought and then I wasn't sure that would be the way to go.

* `mrp_analytic` started in 8.0 creating links between MOs and analytic accounts.
* In 14.0, it added a feature to generate anaytic lines for journal items, which per my understanding lost its meaning to exist in 15.0 when Odoo added analytic_account_id in MO.

The module I'm proposing here kind of replaces the function of the second point, but I wasn't sure if the intended usage was the same (tbh, the design concept the second point wasn't clear to me).

If this module should be a successor of 14.0 `mrp_analytic`, I will try and create another PR following the migration procedure, but I do not have a clear idea at the moment whether the first feature should be migrated, and if yes, how.

Maybe the original authors can advise? @pedrobaeza @dreispt @kjoeldevk

pedrobaeza:
This module should be called `mrp_stock_analytic`, and be independent from `mrp_analytic`. About `mrp_analytic`, it was originally created for having other modules depending on it that creates analytic entries for costs analysis based on the analytic account defined there. Now its goal was perverted for those needing stock accounting and making it dependent of `stock_account`, which for me is not correct.

Anyway, isn't Odoo already tracking MRP production costs other way without analytic?

yostashiro:
OK, thanks Pedro. I'm fine with changing the name to `mrp_stock_analytic`. Any other comments from others?

yostashiro:
> Anyway, isn't Odoo already tracking MRP production costs other way without analytic?

@pedrobaeza Not really aware of it. Do you have a pointer?

pedrobaeza:
Just presentations, but it should be only EE. Not yet analyzed the 16.0.

dreispt:
You could install `mrp` without Accounting, but then you wouldn't have Analytic Accounts available.
So `mrp_analytic` implies having Accounting installed anyway, so that dependency was already there, it is just explicit.
And if you have an Analytic account on an MO, it is kind of expected that it is carried over to the JEs generated by that MO.

I feel that this module conflicts in scope with `mrp_analytic`, and that should be resolved.
I can agree the `mrp_stock_analytic` might be a better name, as it makes the dependencies more explicit that just `mrp_analytic`.

yostashiro:
> I feel that this module conflicts in scope with `mrp_analytic`, and that should be resolved.

@dreispt I've been aware of this.  Can you please go through these comments once again?

https://github.com/OCA/account-analytic/pull/536#issue-1580848505
https://github.com/OCA/account-analytic/pull/536#issuecomment-1433160636
https://github.com/OCA/account-analytic/pull/536#issuecomment-1433169835

> And if you have an Analytic account on an MO, it is kind of expected that it is carried over to the JEs generated by that MO.

I think it was like that in 15.0 core, but not in 16.0 on purpose: https://github.com/odoo/odoo/commit/7e3403068fc3fbc40182b3cfeb80e97a9300e8ff

Not sure if I miss something but I tend to think what #553 tries to do goes against the changes in the core around analytic accounting since https://github.com/odoo/odoo/commit/7e3403068fc3fbc40182b3cfeb80e97a9300e8ff.

dreispt:
I reviewed those comments, and I don't think Odoo 16.0 brings anything new other than replacing the single Analytic Account with an Analytic Distribution.
I expected the `mrp_analytic` migration to handle this, but I could be wrong.
Let me check that.

yostashiro:
@OCA/accounting-maintainers Can this be reopened please?

AungKoKoLin1997 (approved):
LGTM.

dreispt:
I wold like to get this one unblocked.
I see this only adds the `analytic_distribution` field.
So I wonder it this isn't the task `mrp_analytic` was supposed to do.

@yostashiro I can see there is no code on stock oves specifically.
I'm curious how `stock_analytic` module "knows" about the mrp.production analytic distribution to then apply it to the consumption stock moves?
Also, should that also happen on the produced stock moves too?

@pedrobaeza I hear that the goal for `mrp_analytic` is to just set the analytic account.
Should the analytic account be replaced by the `analytic_distribution` field, thus making this change naturally go into the v16 `mrp_analytic` ?

pedrobaeza:
> @pedrobaeza I hear that the goal for mrp_analytic is to just set the analytic account.
Should the analytic account be replaced by the analytic_distribution field, thus making this change naturally go into the v16 mrp_analytic ?

Well, it can evolve to that way, but then the costs should be distributed that way, which can be harder. I think this is more like the analytic account set in the project.

dreispt:
Thank you @pedrobaeza, that makes sense.
And I understand `mrp_analytic` does not make sense anymore, as that field is now added in the core.

AaronHForgeFlow:
Hi @yostashiro  can you please rebase so runboat is rebuilt? thank you

AaronHForgeFlow:
I see the analytic distribution is not passed to the journal items. Is there another module for that feature?

yostashiro:
> I see the analytic distribution is not passed to the journal items. Is there another module for that feature?

@AaronHForgeFlow That's the job of `stock_analytic`, which is a dependency of this module.

AaronHForgeFlow:
@yostashiro , I see. However it is not happening and it is installed in the runboat. Do you think there is  a bug on that module?

Another thing, the analytic distribution is passed from the MO to the stock moves, but only the consumption stock moves, is that intentional?

PR #541, opened 2023-02-22
[16.0][MIG] account_analytic_required

nguyenminhchien:


yostashiro:
@nguyenminhchien Thanks for the PR. Can you please check if this module is really needed in 16.0. I thought the applicability settings of the analytic plan should replace the function of this module.

nguyenminhchien:
> @nguyenminhchien Thanks for the PR. Can you please check if this module is really needed in 16.0. I thought the applicability settings of the analytic plan should replace the function of this module.

Thanks for your feedback, i think we should keep this module. The applicability of the analytic plan is used for validating the analytic distribution. while this module adds a constraint to check if the Analytic is set or not, base on the analytic policy .

StephaneMangin (approved):

simahawk:
> > @nguyenminhchien Thanks for the PR. Can you please check if this module is really needed in 16.0. I thought the applicability settings of the analytic plan should replace the function of this module.
> 
> Thanks for your feedback, i think we should keep this module. The applicability of the analytic plan is used for validating the analytic distribution. while this module adds a constraint to check if the Analytic is set or not, base on the analytic policy .

@alexis-via or @grindtildeath WDYT?

account_analytic_required/models/account.py
```diff
@@ -34,49 +34,31 @@ class AccountAccountType(models.Model):
...
-class AccountAccount(models.Model):
-    _inherit = "account.account"
-
     def _get_analytic_policy(self):
         """Extension point to obtain analytic policy for an account"""
         self.ensure_one()
-        return self.user_type_id.with_company(
-            self.company_id.id
-        ).property_analytic_policy
+        return self.analytic_policy
```
yankinmax, line 40: should we really remove `with_company`?

account_analytic_required/models/account.py
```diff
@@ -34,49 +34,31 @@ class AccountAccountType(models.Model):
...
 
 
 class AccountMove(models.Model):
     _inherit = "account.move"
 
     def _post(self, soft=True):
-        res = super()._post(soft=soft)
+        res = super(AccountMove, self)._post(soft=soft)
```
yankinmax: you don't change `self`, why do we need this change?

account_analytic_required/models/account.py
```diff
@@ -85,27 +67,23 @@ def _check_analytic_required_msg(self):
...
-            self.analytic_account_id or self._has_analytic_distribution()
-        ):
-            analytic_account = (
-                self.analytic_account_id
-                or self.analytic_tag_ids.analytic_distribution_ids[:1]
-            )
+        elif analytic_policy == "never" and (self.analytic_distribution):
+            analytic_account = self.analytic_distribution
+            aaacount_ids = [int(k) for k in analytic_account.keys()]
+            aaacounts = self.env["account.analytic.account"].browse(aaacount_ids)
```
yankinmax: maybe better `analytic_acc_ids` and `analytic_accs` to shorten the vars names

account_analytic_required/models/account.py
nguyenminhchien: i just want to make it explicitly, anyway, i have updated to make no change here

account_analytic_required/models/account.py
nguyenminhchien: updated !

yankinmax (approved):

grindtildeath:
I did some testing on v16.0 (as I hadn't check the new feature before).
I tend to share @yostashiro 's feeling, with the exception that we don't have an option to _always_ require the analytic selection and the mandatory applies for posted moves (but IMO there isn't much sense to strictly require a definition while the record is still in draft).
The issue I can see with the migrated module is that such a configuration can then either be done through the account account model or the analytic plan.
I may have missed something and there might still be a good reason to migrate as is, but I don't see it yet and would appreciate an explanation with proper use case describing what we cannot do with the standard.

account_analytic_required/README.rst
```diff
@@ -0,0 +1,100 @@
...
+You have the choice between 4 policies : *always*, *never*, *posted moves* and *optional*.
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
```
rafaelbn, line 38: - Go to Accounting -> Configuration and activate: Analytic Accounting
- Go to... (I don't know hoy to continue) 😢

rafaelbn (changes requested):
Please could you improve readme?

I'm unable to find account type to configure this module 😄 

Hoy did you test it? I'm trying it but no success

Thank you!

account_analytic_required/README.rst
nguyenminhchien: I don't know why you don't see these lines:
```
Example:

If you want to have an analytic account on all your *expenses*,
set the policy to *always* for the account of type *expense*.
If you try to save a journal items with an account of type *expense*
without analytic account, you will get an error message.
```
anw, because "Account type" has been removed, the configuration will be done on the Account via menu: **Invoicing > Configuration > Accounting > Chart of Accounts**.

rafaelbn:
We should review deeper, going account by account (hundreds) activating this is not usable.

rafaelbn:
- [x] @nguyenminhchien we should adapt the required to the new account type rework in https://github.com/odoo/odoo/commit/26b2472f4977ccedbb0b5ed5f08be2c04313fd21

rafaelbn:
Hello @nguyenminhchien , I'm testing again it not easy to configure. In Spain we have by default 132 Expense accounts and 121 Income accounts, repeat 😄 **by default (legal ones)**

rafaelbn (commented):
Dear all! I guess this module is not needed any more so migration is not needed.

Rationale

- [x] Analytic account has mandatory the Analytic plan
- [x] You can set Analytic plan to be mandatory, so this is the same of this module.

Video: https://www.loom.com/share/2c6928e3380a4ae8a1802a277065120d

Other opinions from you and @OCA/accounting-maintainers are welcome

In my side should be closed and use standard

sbidoul:
> You can set Analytic plan to be mandatory, so this is the same of this module

@rafaelbn are you sure this constraint is applied in all circumstances? I think @alexis-via mentioned there are cases where the validation is not done by Odoo.

rafaelbn:
> @rafaelbn are you sure this constraint is applied in all circumstances? I think @alexis-via mentioned there are cases where the validation is not done by Odoo.

- If you set analytic plan as mandatory it happens.
- If you set as not mandatory then no

Where @alexis-via said that? Twitter? Is there a link?

In all circumstances?

- Do you mean in maybe other views like assets, and so on?

The module doesn't hurt. I mean, if migrated and installed works with Plans

Know that you comment, maybe using some other OCA modules doesn't work as Odoo don't think about other modules.

I'm not blocking at all! 😄 ❤️ just my test to decrease modules when it is possible.

Best regards!

rafaelbn (approved):
Functionally tested module works 👍 

I think functionality is in core with planes but maybe I missed something.

Up to you! 😄

sbidoul:
@rafaelbn I was referring to [this tweet](https://twitter.com/alexisdlattre/status/1633606705304018946). I've not investigated myself, though so I don't know the details.

account_analytic_required/tests/test_account_analytic_required.py
```diff
@@ -188,13 +161,3 @@ def test_posted_ok(self):
         move = line.move_id
         move.action_post()
         self.assertEqual(move.state, "posted")
-
-    def test_exception_policy_for_analytic_distribution(self):
```
leemannd: Any reason to drop this test? Isn't it still in the scope of the module?

nguyenminhchien: Analytic Distribution is Analytic, no need to make this test anymore.

StephaneMangin (approved):

account_analytic_required/models/account.py
```diff
@@ -0,0 +1,105 @@
...
+    analytic_policy = fields.Selection(
+        selection=[
+            ("optional", "Optional"),
+            ("always", "Always"),
+            ("posted", "Posted moves"),
+            ("never", "Never"),
+        ],
+        string="Policy for analytic account",
+        default="optional",
+        company_dependent=True,
```
alexis-via, line 21: ARF !!! Not on account.account model which is always attached to a specific company !!!

PR #543, opened 2023-02-27
[16.0][IMP] stock_analytic: more granularity on applicability judgment

yostashiro:
Before this commit, analytic distribution would be validated only for outgoing moves, which was not ideal in view of a module such as purchase_stock_analytic.

This commit adds more flexibility to the applicability judgment by adding the stock picking type field to the analytic applicability model, removing the assumption of validating outgoing moves only.

@qrtl


AungKoKoLin1997 (approved):
Function tested in runboat.
It worked as expected.

PR #554, opened 2023-05-09
[16.0][ADD] stock_lot_analytic

AungKoKoLin1997:
This module adds Analytic Distribution field in stock lot to be able to get analytic information when lots/serials is created in purchase receipts.

@qrtl


AungKoKoLin1997:
I considered this module is not totally generic to maintain in OCA repo.

PR #561, opened 2023-06-05
[16.0][IMP] purchase_analytic: add default value in order line

AungKoKoLin1997:
This PR take default values in order line if there is analytic_distribution in purchase order.

@qrtl


yostashiro (approved):
Functional test.  Although this adds a new dependency `base_view_inheritance_extension`, it makes the approach consistent with `stock_picking_analytic` (following code is from 14.0).

https://github.com/OCA/account-analytic/blob/9e7039ea1f5a582fc100f45944f13ce3452f33b3/stock_picking_analytic/views/stock_picking.xml#L28-L34

AaronHForgeFlow (approved):

PR #563, opened 2023-06-13
[IMP] purchase_analytic: fix project_id being erase after selecting p…

JuanyDForgeflow:
Old behavior: When selecting a new product the analytic account gets erased
Current behavior: the new added line gets the analytic account assigned


AaronHForgeFlow (commented):
Works good. 10.0 branch CI won't be fixed unfortunately :cry:

PR #565, opened 2023-06-16
[16.0][ADD] analytic_mixin_analytic_account

AungKoKoLin1997:
This module adds a many2many analytic accounts field to the analytic mixin, calculated from the value of the analytic distribution, in an effort to mitigate the difficulties associated with the absence of these fields directly on the model inheriting the mixin.

@qrtl


yostashiro (commented):
@AungKoKoLin1997 Can you please add tests.

yostashiro (approved):
Code review and functional test.

ghost (commented):
I think it would be better to have the fields stored 

On the other hand, It would be great to have it integrated at some point with others modules like this one - https://github.com/OCA/account-financial-reporting/blob/16.0/account_financial_report/models/account_move_line.py - to avoid having several logic in the OCA

AungKoKoLin1997:
> I think it would be better to have the fields stored

I don't see the reason to make the fields stored at the moment.

> On the other hand, It would be great to have it integrated at some point with others modules like this one - https://github.com/OCA/account-financial-reporting/blob/16.0/account_financial_report/models/account_move_line.py - to avoid having several logic in the OCA

Yes, I think so. After this module is merged, we can make adjustment to other modules that want to have analytic accounts by just inheriting this module.

AaronHForgeFlow:
This is really nice. As long as many modules will depend on this one I think it is better to introduce it as LGPL, that way we don't have to modify the license in the modules that inherit from this one.

yostashiro:
Yes, let's make it LGPL.

@jdidderen-noviat Do you have some specific inconveniences in mind if we do not make the field stored?

ghost:
> @jdidderen-noviat Do you have some specific inconveniences in mind if we do not make the field stored?

If I take the example of the account_financial_report:

- The analytic_account_ids field is used in search domains (https://github.com/OCA/account-financial-reporting/blob/89144fe180ca526d3382ee7bb159b99e39602c68/account_financial_report/report/general_ledger.py#L278) 
- In some cases, financial reports search through thousands of accounting entries, and if the values have to be calculated for each record, in my opinion, it's a loss in performance.

AungKoKoLin1997:
@jdidderen-noviat Thanks for the pointer. I overlooked a concept used in reporting.

AungKoKoLin1997:
Creating stored fields is challenging in this module. Simply adding the store attribute results in overly long table names, leading to errors. I encountered the following error:
`odoo.http: Table name 'account_analytic_account_account_analytic_distribution_model_rel' is too long.`

To address this, I tried specifying the relation attribute as shown below:

```
analytic_account_ids = fields.Many2many(
    "account.analytic.account",
    relation="analytic_mixin_account_rel",
    column1="mixin_id",
    column2="account_id",
    compute="_compute_analytic_account_ids",
    string="Analytic Accounts",
    store=True,
    help="Analytic accounts computed from analytic distribution.",
)
```
However, this introduced another problem. Since several models inherit from analytic.mixin, it results in a conflict, as evidenced by the error:
`TypeError: Many2many fields account.reconcile.model.line.analytic_account_ids and account.analytic.distribution.model.analytic_account_ids use the same table and columns.`

Any suggestions?

yostashiro:
A quick workaround I can think of is to shorten the field name to something like `analytic_acct_ids` which should keep the table name within 63-char limit.

AungKoKoLin1997:
> A quick workaround I can think of is to shorten the field name to something like `analytic_acct_ids` which should keep the table name within 63-char limit.

@yostashiro Unless I'm mistaken, the field name doesn't influence the generation of the table name for a Many2many relationship in Odoo. The format typically follows the pattern `model_1_name_model_2_name_rel`.

yostashiro:
Oh yes, that's right. Sorry.

AungKoKoLin1997:
I've chosen to keep non-stored fields in this module for two main reasons:

- **Technical**: Due to the reasons previously mentioned. (We might, however, force stored fields in each inherited model.)
- **Functionality**: To prevent excessive strain on existing records when we use stored fields. For instance, other models depend on      `analytic.mixin`, and we would need to handle all of them using hooks.

norlinhenrik (approved):

AungKoKoLin1997:
@jdidderen-noviat Do you have any suggestions or concern regarding these comments (https://github.com/OCA/account-analytic/pull/565#issuecomment-1742393181, https://github.com/OCA/account-analytic/pull/565#issuecomment-1744047702)

norlinhenrik:
> I've chosen to keep non-stored fields in this module for two main reasons:
> 
> * **Technical**: Due to the reasons previously mentioned. (We might, however, force stored fields in each inherited model.)
> * **Functionality**: To prevent excessive strain on existing records when we use stored fields. For instance, other models depend on      `analytic.mixin`, and we would need to handle all of them using hooks.

I support keeping non-stored fields by default, and force stored field in the inherited model when needed.

Can we merge?

norlinhenrik:
After merging, please review #630 which depends on this PR.

norlinhenrik:
@sbidoul Would you like to review?

AungKoKoLin1997:
@OCA/account-analytic-maintainers 
Could you please review this PR?

norlinhenrik:
What do you think about computing 2 more fields: `analytic_plan_ids` & `analytic_plan_names` ?
If we can merge this PR, I can open a new PR with this improvement.

AaronHForgeFlow (approved):
LGTM

norlinhenrik:
> On the other hand, It would be great to have it integrated at some point with others modules like this one - https://github.com/OCA/account-financial-reporting/blob/16.0/account_financial_report/models/account_move_line.py - to avoid having several logic in the OCA

I have made #673 as an alternative to this PR, and [#1207 for account_financial_report](https://github.com/OCA/account-financial-reporting/pull/1207).

analytic_mixin_analytic_account/models/analytic_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+        string="Analytic Accounts",
+        help="Analytic accounts computed from analytic distribution.",
+    )
+    analytic_account_names = fields.Char(
+        compute="_compute_analytic_account_ids",
+        help="Comma-separated analytic account names, in case it is useful to be "
+        "included in the exported data.",
+    )
+
+    def _compute_analytic_account_ids(self):
```
sbidoul, line 24:
```suggestion
    @api.depends("analytic_distribution")
    def _compute_analytic_account_ids(self):
```

analytic_mixin_analytic_account/models/analytic_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+        analytic_account = self.env["account.analytic.account"]
```
sbidoul:
```suggestion
        AnalyticAccount = self.env["account.analytic.account"]
```

or `analytic_account_model`

sbidoul (commented):
A couple of comments.

For my use cases (record rules) I'd need to have this field stored too, so not sure I will be able to use this module as is.

analytic_mixin_analytic_account/models/analytic_mixin.py
AungKoKoLin1997: I updated the code. Thanks!

analytic_mixin_analytic_account/models/analytic_mixin.py
AungKoKoLin1997: I update the code to use as `analytic_account_model`.

analytic_mixin_analytic_account/models/analytic_mixin.py
```diff
@@ -0,0 +1,34 @@
...
+
+from odoo import api, fields, models
+
+
+class AnalyticMixin(models.AbstractModel):
+    _inherit = "analytic.mixin"
+
+    analytic_account_ids = fields.Many2many(
+        "account.analytic.account",
+        compute="_compute_analytic_account_ids",
```
yostashiro, line 12: To handle archived analytic accounts as well.
```suggestion
        compute="_compute_analytic_account_ids",
        context={"active_test": False},
```

analytic_mixin_analytic_account/models/analytic_mixin.py
```diff
@@ -0,0 +1,34 @@
...
+    @api.depends("analytic_distribution")
+    def _compute_analytic_account_ids(self):
+        analytic_account_model = self.env["account.analytic.account"]
+        for rec in self:
+            if not rec.analytic_distribution:
+                rec.analytic_account_ids = analytic_account_model
+                rec.analytic_account_names = False
+                continue
+            account_ids = [int(key) for key in rec.analytic_distribution.keys()]
+            rec.analytic_account_ids = analytic_account_model.browse(account_ids)
```
yostashiro: To be a bit more concise and also to handle archived analytic accounts.
```suggestion
        for rec in self:
            if not rec.analytic_distribution:
                rec.analytic_account_ids = False
                rec.analytic_account_names = False
                continue
            account_ids = [int(key) for key in rec.analytic_distribution.keys()]
            rec.analytic_account_ids = [(6, 0, account_ids)]
```

yostashiro (changes requested):

yostashiro (approved):

kanda999 (approved):
Functional test:  It works as expected.

norlinhenrik:
@kanda999 Thank you!

@sbidoul Would you like to merge?

AungKoKoLin1997:
@AaronHForgeFlow Can you please merge?

AaronHForgeFlow:
@AungKoKoLin1997 I rather like to have @sbidoul approval. There is some discussion about making the fields stored and I am not  sure if there is a final agreement.

sbidoul (commented):
For the use case mentioned, it should be ok and the code looks good to me.
I don't have a use for this module myself, though. So don't wait on me :)

analytic_mixin_analytic_account/models/analytic_mixin.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @api.depends("analytic_distribution")
+    def _compute_analytic_account_ids(self):
+        for rec in self:
+            if not rec.analytic_distribution:
+                rec.analytic_account_ids = False
+                rec.analytic_account_names = False
+                continue
+            account_ids = [int(key) for key in rec.analytic_distribution.keys()]
+            rec.analytic_account_ids = [(6, 0, account_ids)]
```
sbidoul: Use Command.set? Or browse(analytic_account_ids)?

norlinhenrik: `record.analytic_account_ids = Command.set(account_ids)`

AungKoKoLin1997: I used Command.set.

kanda999 (approved):

norlinhenrik (approved):

yostashiro (approved):

AaronHForgeFlow (approved):
Fair enough. Code review LGTM

PR #570, opened 2023-07-03
[13.0][FIX]product_analytic: ensure move type in dict

GuillemCForgeFlow:
There may be a case where the move type is not contained in the global dict keys. Perhaps we need to ensure that we have the type in one of the dict keys.

cc @ForgeFlow


JordiMForgeFlow (approved):

AaronHForgeFlow:
Fasttrack

/ocabot merge patch

Issue #551, opened 2023-03-29
How the OCA will deal with the disappearance of analytic tags


rafaelbn:
Hello all,

Odoo deletes and loses all analytics tags in Odoo 16

OCA has lot of modules over this model

1. account.analytic.group-> account.analytic.plan
2. account.analytic.account -> account.analytic.account
3. account.analytic.tag -> Deletes (All data is lost)
4. account.analytic.default` -> Deletes (All data is lost)

I would like to discuss what is on @OCA/accounting-maintainers and general with contributors what should be done in:

- [ ] Module migrations
- [ ] OpenUpgrade

### Analytic Tags should be maintained just as tags

**Analytic tags had 2 functions:**

- For labeling analtytic lines + account moves
- Making analytical distributions

i.e. it does 2 things

**Analytic plans have 1 function:**

- Multiple distributions in different plans

IMHO: The new analytical plans are the new distributions, but the **analytical tags must be maintained by removing the distribution function**.

Thank you
Regards

kharkot:
Hello ,

In that subject I have logged ticket to Odoo - you can track my correspondence https://www.odoo.com/my/tasks/3173120?access_token=f9c64ccd-b1a9-4e36-905c-29a082c14426  

For while functionality (only field on account.move) was removed but....

rafaelbn:
### Analytic Tags should be maintained just as tags

**Analytic tags had 2 functions**

- For labeling analtytic lines + account moves
- Making analytical distributions

i.e. it does 2 things

**Analytic plans have 1 function:**

- Multiple distributions in different plans

IMHO: The new analytical plans are the new distributions, but the **analytical tags must be maintained by removing the distribution function**.

kharkot:
@rafaelbn 
Recommendation from Odoo:

`Since you are in Odoo 15, I would recommend you to not upgrade to Odoo 16 and wait for Odoo 17 which will have its analytic tags way of working improved.

Once again, we apologize that the changes made in Odoo 16 do not suit you and your business.
We are already working on it to improve it in Odoo 17.`

Shide:
I've found a problem when migration has been done via Odoo.
The relation with `account.move.line` and `account.analytic.tag` has been stored on the table `account_analytic_tag_account_move_line_rel`.

When Odoo is migrating, removes the table from the database, and it causes losing all the data.
[...]

PR #572, opened 2023-07-11
[16.0][ADD] account_analytic_tag

Yadier-Tecnativa:
Restore account.analytic.tag
Solve the issue https://github.com/OCA/account-analytic/issues/551

@Tecnativa TT44062


Yadier-Tecnativa:
ping @sergio-teruel , @pedrobaeza

account_analytic_tag/models/account_analytic_tag.py
```diff
@@ -0,0 +1,53 @@
+from odoo import fields, models
+
+
+class AccountAnalyticDistribution(models.Model):
```
pedrobaeza: This one is not needed anymore. You only need the tag

account_analytic_tag/models/account_analytic_tag.py
```diff
@@ -0,0 +1,53 @@
...
+    name = fields.Char(string="Analytic Tag", index=True, required=True)
+    color = fields.Integer("Color Index")
+    active = fields.Boolean(
+        default=True,
+        help="Set active to false to hide the Analytic Tag without removing it.",
+    )
+    active_analytic_distribution = fields.Boolean("Analytic Distribution")
+    analytic_distribution_ids = fields.One2many(
+        "account.analytic.distribution", "tag_id", string="Analytic Accounts"
+    )
```
pedrobaeza: Remove these 2 fields. We are only wanting the tag part.

account_analytic_tag/models/account_analytic_tag.py
```diff
@@ -0,0 +1,53 @@
...
+    company_id = fields.Many2one("res.company", string="Company")
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
```
pedrobaeza: This should go into a separate file.

account_analytic_tag/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module adds *account analytic tags* on accounts.
```
pedrobaeza:
```suggestion
This module restores the *account analytic tags* as a method for categorizing analytic entries, selectable from the invoices, and transferred to them on publishing.
```

account_analytic_tag/readme/CREDITS.rst
pedrobaeza, line 1: Don't add empty files.

account_analytic_tag/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+Analytical tags.
```
pedrobaeza, line 1: Please elaborate this a bit more, and put in `CONFIG.rst` the part that belongs there.

account_analytic_tag/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Yadier Quesada - Tecnativa <yadier.dequesada@tecnativa.com>
```
pedrobaeza: This is not the form we put it:

```suggestion
* `Tecnativa <https://www.tecnativa.com>`__:

  * Yadier Quesada
```

pedrobaeza (changes requested):
This is missing a lot of things:

- Analytic tags on invoices.
- Transfer such tags on the created analytic entries when publishing the invoices.

Yadier-Tecnativa:
All suggestions were made, please review @pedrobaeza

Yadier-Tecnativa:
ping @pilarvargas-tecnativa , @carlosdauden Please review

rafaelbn:
Thank you! Great contribution! I will review this Friday!

pepetreshere:
@Yadier-Tecnativa @pedrobaeza nice PR!, would you please drop some light on why do you remove the analytic_distribution_ids? The analytic_distribution_ids allowed to have a "predefined" percentage for each analytic account so that it is easier for and end-user to quickly select a set of "predefined" percentages. This is not covered by standard Odoo in v16.0 not even by using the new analytic plans. Thanks in advance for any light

pedrobaeza:
Because this feature is already covered by standard v16.

rafaelbn:
Hello! Really thank you very much for this contribution.

The implementation with this PR of analytic tags is different from Odoo 15 and previous. In this moment functionality of this module duplicated the function of Analytic Plans.

-------------------

- [ ] Analytic tags cannot have analytic account mandatory as they are independent, in fact don't need any relation.

You create analytic tags for many analytic accounts that will be created in the future (like projects)

Important: When the mind behind this decision in Odoo deleted the tags, he/she thought that all Odoo customers worldwide where using tags for distributions! But not, just tag. See

imagen

I have recorded a video and created an excel 😄 see:

- https://github.com/OCA/account-analytic/issues/551#issuecomment-1672273501

This PR should be able to reproduce that use case like in Odoo 15, 14, 13, 12, 11, etc.

--------------------

- [ ] There is not menu to add, see or edit analytic tags

imagen

---------------------

- [ ] The setting option to activate is missing

imagen


----------------------

- [ ] We cannot set analytic tags in invoice lines, the field is not available

----------------------


Let me know please if I can help you further

Regards,
Rafael

account_analytic_tag/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+Example:
+
+If you want to use analytic account tags ...
+
+Usage
+=====
+
+Analytical tags.
+This function allows you to create formulas to establish the distribution of prices
```
rafaelbn: @Yadier-Tecnativa this is not the function of analytic tags

account_analytic_tag/README.rst
```diff
@@ -0,0 +1,96 @@
...
+between analytical accounts. This means that when we create an analytics tag,
+we must establish an Analytical Account to which it is directed and the monetary percentage.
```
rafaelbn: As commented previously not this is not. Please check the video I recorded 🙏🏼 😄

rafaelbn (changes requested):

sbidoul:
Thanks for this work. 

I think there are many different ways people were using analytic tags with other versions.

Would this make sense to you that we split the feature with first a base module that just provides the `account.analytic.tag` model, then additional modules that make use of them. Such as using tags on journal items, or settings tags on analytic accounts (as was possible until v11) ?

account_analytic_tag/models/account_analytic_tag.py
```diff
@@ -0,0 +1,17 @@
...
+    name = fields.Char(string="Analytic Tag", index=True, required=True)
+    color = fields.Integer("Color Index")
+    active = fields.Boolean(
+        default=True,
+        help="Set active to false to hide the Analytic Tag without removing it.",
+    )
+    company_id = fields.Many2one("res.company", string="Company")
+    account_analytic_id = fields.Many2one(
+        "account.analytic.account", string="Analytic Account", required=True
+    )
```
sbidoul, line 22: This m2o field seems wrong ?

pedrobaeza:
> Would this make sense to you that we split the feature with first a base module that just provides the account.analytic.tag model, then additional modules that make use of them. Such as using tags on journal items, or settings tags on analytic accounts (as was possible until v11) ?

I don't see many sense on splitting model definition from account.move.line links at least. No need to extra dependencies (all are in `account`). It's complicating the module tree, even increasing maintenance burden, while the features are code are very simple to be maintained even if packed.

pedrobaeza:
FYI, I will overtake this PR and finish it.

sbidoul:
> I don't see many sense on splitting model definition from account.move.line links at least

In our use case we need the tags but not the link with `account.move.line` because the tags are set on analytic accounts (it was possible until v11). 

So I don't mind if the link with move lines is in the same module, but then we also need the group to show the analytic tags on move lines only to selected users.

In general I prefer small modules with restricted scope because they are easier to migrate. But I'm not blocking on this, especially if you do the work :).

> FYI, I will overtake this PR and finish it.

@pedrobaeza thanks! Do you know approximately when you'll work on that?

pedrobaeza:
Probably next week.

rafaelbn:
> I don't see many sense on splitting model definition from account.move.line links at least. No need to extra dependencies (all are in account). It's complicating the module tree, even increasing maintenance burden, while the features are code are very simple to be maintained even if packed.

I agree 👍🏼 , thank you!


> In our use case we need the tags but not the link with `account.move.line` because the tags are set on analytic accounts (it was possible until v11).

@sbidoul then you set `account.analytic.tag` in `account.analytic.account` and you don't want that this tags are propagated to `account.move.line` ? I don't catch it. Maybe we can talk about this in other moment (for me it is interesting)


> but then we also need the group to show the analytic tags on move lines only to selected users.

Interesting too. I guess the users with permissions in analytic with an extra security group can.


> FYI, I will overtake this PR and finish it.

Awesome!

Thank you ver much.

victoralmau:
The following changes have been done:

- [x] Added checkbox at the billing configuration level for analytical tags.
- [x] Removed the active / inactive check in the labels.
- [x] Color field added in tree and form view.
- [x] Define the corresponding groups in the `company_id` field.
- [x] Analytic account field is not mandatory now to create labels.
- [x] The process of creating analytical items is simplified; now only analytical tags are set in the corresponding analytical items (those with the same or no analytical account defined in the label).
- [x] Create base tests to be able to use them in other modules that extend this functionality.
- [x] Add more use cases in tests.

account_analytic_tag/README.rst
```diff
@@ -0,0 +1,96 @@
...
+**Table of contents**
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To manage analytic account tags using this module, you need to:
+
```
rafaelbn, line 40: Go to Accounting -> Configuraton -> Settings -> Activate "Analytic Tags"

account_analytic_tag/models/account_analytic_tag.py
sbidoul: Can you add a help text that explains the purpose of this `account_analytic_id` field? Perhaps a more specific label than `Analytic Account` would help too.

account_analytic_tag/models/account_move_line.py
```diff
@@ -0,0 +1,25 @@
...
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    analytic_tag_ids = fields.Many2many(
+        comodel_name="account.analytic.tag",
+        store=True,
+        string="Analytic Tags",
+        readonly=False,
+    )
```
sbidoul:
```suggestion
    analytic_tag_ids = fields.Many2many(
        comodel_name="account.analytic.tag",
        string="Analytic Tags",
    )
```

account_analytic_tag/views/account_analytic_line_views.xml
```diff
@@ -0,0 +1,24 @@
+<odoo>
+    <!-- Inherit Form View to Modify it -->
+    <record id="view_account_analytic_line_tree_inherit_tags" model="ir.ui.view">
+        <field name="name">account.analytic.line.tree.inherit.tags</field>
+        <field name="model">account.analytic.line</field>
+        <field name="inherit_id" ref="analytic.view_account_analytic_line_tree" />
+        <field name="arch" type="xml">
+            <field name="account_id" position="after">
+                <field name="tag_ids" widget="many2many_tags" optional="hide" />
+                <field name="product_uom_category_id" invisible="1" />
```
sbidoul: This `product_uom_category_id` field seems unrelated to this module ?

account_analytic_tag/views/account_analytic_line_views.xml
```diff
@@ -0,0 +1,24 @@
+<odoo>
+    <!-- Inherit Form View to Modify it -->
+    <record id="view_account_analytic_line_tree_inherit_tags" model="ir.ui.view">
+        <field name="name">account.analytic.line.tree.inherit.tags</field>
+        <field name="model">account.analytic.line</field>
+        <field name="inherit_id" ref="analytic.view_account_analytic_line_tree" />
+        <field name="arch" type="xml">
+            <field name="account_id" position="after">
+                <field name="tag_ids" widget="many2many_tags" optional="hide" />
```
sbidoul: Can you make this field visible only to the `account_analytic_tag.group_analytic_tags` group?

account_analytic_tag/views/account_analytic_line_views.xml
```diff
@@ -0,0 +1,24 @@
...
+            </field>
+        </field>
+    </record>
+    <record id="view_account_analytic_line_form" model="ir.ui.view">
+        <field name="name">account.analytic.line.form.inherit.tags</field>
+        <field name="model">account.analytic.line</field>
+        <field name="inherit_id" ref="analytic.view_account_analytic_line_form" />
+        <field name="arch" type="xml">
+            <field name="account_id" position="after">
+                <field name="tag_ids" widget="many2many_tags" />
```
sbidoul: Make it visible to the `account_analytic_tag.group_analytic_tags` group only.

account_analytic_tag/views/account_analytic_tag_views.xml
```diff
@@ -0,0 +1,77 @@
...
+                Add a new tag
+            </p>
+        </field>
+    </record>
+    <menuitem
+        id="account_analytic_tag_menu"
+        name="Analytic Tags"
+        parent="account.menu_analytic_accounting"
+        action="account_analytic_tag_action"
+        groups="group_analytic_tags"
```
sbidoul: Can we make this menu visible to the `account.group_account_manager` group instead. 

So accounting administrators can configure tags, but only users in the `account_analytic_tag.group_analytic_tags` can use them. This is important so we can use analytic tags just to categorize analytic accounts (in #585) but not show analytic tags on invoice/move lines.

sbidoul (commented):
Thanks for this work. Here are a few comments.

victoralmau:
> Thanks for this work. Here are a few comments.

Changes done.

sbidoul:
Thanks! Code looks good to me. 

Do you know why the tag colors don't show on the invoice screen?

http://oca-account-analytic-16-0-pr572-a0742ba028bb.runboat.odoo-community.org/web?debug=1#id=18&cids=1&menu_id=115&action=232&active_id=1&model=account.move&view_type=form

image

image

account_analytic_tag/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_account_analytic_tag,account.analytic.tag,model_account_analytic_tag,analytic.group_analytic_accounting,1,1,1,1
```
sbidoul: @victoralmau oh, something needs fixing here too.

It should be readonly for `account_analytic_tag.group_analytic_tags` (not `analytic.group_analytic_accounting`, BTW).
And 1,1,1,1 for `account.group_account_manager` since managing the available tags is an accounting administration feature.

victoralmau:
Changes done.

sbidoul (approved):
Thank you so much @victoralmau !

rafaelbn:
I will finish my review **today**! This evening 😄

pedrobaeza:
I'm thinking on reintroducing the distributions on tags as others have commented, as a way of "templating" this without defining a distribution model. What do you think about it?

victoralmau:
> I'm thinking on reintroducing the distributions on tags as others have commented, as a way of "templating" this without defining a distribution model. What do you think about it?

This is a very good point.

Previously (up to v15) we all modeled the analytical distribution with labels, but now in v16 the concept of distributions has been introduced.

Advantages / disadvantages of each option.

**Option A: Keep the distributions in the labels**.

- [x] Disadvantage: Would duplicate functionality with distributions.
- [x] Advantage: It would work the same as in v15 and before.
- [x] Advantage: It will not be necessary to define a distribution on the invoice lines.
- [x] Disadvantage: It will be necessary to create the analytical entries when there is no distribution defined. What happens if there are labels and analytical distribution?
- [x] Advantage: No need to define distributions in OCA modules that add analytics (e.g. `project_stock`).

**Option B: Leave the analytical tags without distributions**.

- [x] Disadvantage: Having to define the concept of distributions in OCA modules that add analytics (e.g. `project_stock`).
- [x] Advantage: Not having to add code to create analytical entries without distributions.

Maybe option A is the most reasonable, what do others think?

sbidoul:
> I'm thinking on reintroducing the distributions on tags as others have commented, as a way of "templating" this without defining a distribution model. What do you think about it?

@pedrobaeza  Can this be done in another module, or in a second step after merging this PR?

sbidoul:
> Disadvantage: Would duplicate functionality with distributions

To me this justifies doing another analytic distributions mechanisms in a different module.

victoralmau:
> > Disadvantage: Would duplicate functionality with distributions
> 
> To me this justifies doing another analytic distributions mechanisms in a different module.

Ok, thinking about this again, I think the best thing to do would be:

`account_analytic_tag_distribution`

- A checkbox is added in Invoicing > Configuration to enable distribution at analytic tag level.
- It hides the Analytic Distribution + Analytic Distribution columns menu if the checkbox is checked.
- The possibility to define the analytical distribution on tags is "restored".
- When defining an analytical label on an invoice line, what is defined on the labels will be defined as distribution.

In this way, I think it avoids "duplicating" the operation of the analytical distributions and could be used in one way or another.

pedrobaeza (approved):
OK about a separated module for distributions.

rafaelbn:
Hello,

Thank you for this great work @victoralmau 

After review, I see this (let me know your opinions):

- [ ] In Invoice lines, set Analytic Tags after Analytic

imagen



- [ ] Go to Accounting -> Accounting -> Journal Items , Then we cannot group by "Analytic Account"

imagen



- [ ] Go to Accounting -> Accounting -> Management -> Analytic Items, change label "Tags" by "Analytic Tags" for better fields order (as alphabetical) and understanding

imagen


In a separate comment I can give my personal opinion about distributions.

Thank you very much! ❤️

rafaelbn (approved):
Hello,

Thank you for this great work @victoralmau 

After review, I see this (let me know your opinions):

- [x] In Invoice lines, set Analytic Tags after Analytic

imagen



- [ ] Go to Accounting -> Accounting -> Journal Items , Then we cannot group by "Analytic Account"

imagen



- [ ] Go to Accounting -> Accounting -> Management -> Analytic Items, change label "Tags" by "Analytic Tags" for better fields order (as alphabetical) and understanding

imagen


In a separate comment I can give my personal opinion about distributions.

Thank you very much! ❤️

rafaelbn:
Hello, this is my opinion about a new module for distributions.

The most important writing this down here is that I don't want to be right, I am aware that there are many different types of Odoo implementations and different types of companies and clients. I don't want to convince anyone. I just want to provide a vision that may help someone else make their own decisions. In fact, **I really like reading your opinions and decisions** and they make me wonder why and what I can learn from it.

So...

If we in OCA reintroduce analytic distributions in tags with are confusing users further. As for every one in Odoo distributions are done in Analytic Plans (`account.analytic.plan`). This re-introducing feature for me, maybe could be decision prompted by the effort to understand the change and the technical dependencies of other modules.

I'm not opposed to anything 😄, this is an open community, but I do think that in addition to having a **separate module** we should make a separate repository like https://github.com/OCA/account-analytic-tags-distribution . Because we must keep an open scenario where contributors extend Odoo using Odoo analytics distributions with their analytics Plans.

In other case, this will be really confusing when developing modules around distributions in Analytic Plans like Odoo Core and Analytic Tags as older versions. In Odoo 18 nobody (new contributors) will remember about why are Distributions in Tags talking around product vision.

You have done a really nice analysis 👍🏼 take this point of view just if it gives you more perspective:

Perfect migrations from Odoo 15 to Odoo 16 for me is like:

- If `account.analytic.tag` has fields distribution "False" Then maintain it in `account.analytic.tag`
- If `account.analytic.tag` has fields distribution "True" Then migrate it to `account.analytic.plan`


Opining and introducing more points about this **great** 🥇 pros and cons analysis:

> Advantages / disadvantages of each option.
> 
> **Option A: Keep the distributions in the labels.**
> 
> - [x] Disadvantage: Would duplicate functionality with distributions.
> - [ ] Advantage: **It would work the same as in v15 and before.**

For me this is a **disadvantage** as you are separating important concepts in Odoo Core. Imagine you decide in Odoo 13 to maintain account.invoice model as a similar example.

> - [ ] Advantage: It will not be necessary to define a distribution on the invoice lines.
> - [x] Disadvantage: It will be necessary to create the analytical entries when there is no distribution defined. What happens if there are labels and analytical distribution?
> - [ ] Advantage: No need to define distributions in OCA modules that add analytics (e.g. project_stock).

For me this is not a Advantage as **you don't need to** define ~~distributions~~ analytic plans. The is always one by default that works in the same way as I you have in Odoo 15 only analytic accounts without distributions. Really I don't see the problem with `project_stock`
Are you talking around UX introducing invoices? I would really love to go deep in this problem that I don't see.

- [x] Disadvantage: You are missing the amazing new great functionality around **Applicability** in Analytic Plans with all the automation rules. https://www.odoo.com/documentation/16.0/applications/finance/accounting/reporting/analytic_accounting.html#analytic-accounting-analytic-plans

> **Option B: Leave the analytical tags without distributions.**
> 
> - [ ] Disadvantage: Having to define the concept of distributions in OCA modules that add analytics (e.g. project_stock).

As said before, you don't have to define concept of distributions. It's not a must, it is an option. Analytic Plans are just plans of analytic account, grouped, that maybe you you may want to distribute among them or not.

> 
> - [x] Advantage: Not having to add code to create analytical entries without distributions.
> 
> Maybe option A is the most reasonable, what do others think?

If you think is needed is OK. I think all the time about have we have (a lot) to not repeat my self (DRY)

1. "Analytic Distribution Models" `account.analytic.distribution.model` are not mandatory as they are rules for automations similar to previous "Analytic Defaults Rules" `account.analytic.default`
2. "Analytic Plans" `account.analytic.plan` are like "Dimensions", they group Analytic accounts and give you the possibility to make distributions or not.
3. Analytic Accounts, are values of the Plans (dimensions) with a direct relations
4. Analytic Tags `account.analytic.tag` in this PR are another level of analysis, for other values and dimensions that are not related so you can assign freely.


My best!, sorry for the length! (my fault)

💯 🥇 👍🏼 

Thank you very much for this work! 👍🏼

victoralmau:
Changes done:
- [x]  In Invoice lines, set Analytic Tags after Analytic
- [x] Change `tag_ids` position in analytical items (same as in v15).

Other comments:
- There is no analytical account field in the Journal Items, they are built according to the analytical distribution.
- Tags is the correct name of the field (same as in v15) in the analytical items.
- The configuration section of the readme already indicates the activation of the "Analytic Tag" checkbox.

pedrobaeza:
@victoralmau put `Co-Authored-By: ...` in the commit for your co-authorship.

rafaelbn:
/ocabot merge nobump

Fantastic! You are awesome! 💚

PR #574, opened 2023-07-12
[15.0][FIX] account_tag_dimension: fix permission sudo for allow modify ir.model

Saran440:
Step to error:
1. install module `account_tag_dimension`
2. create user accounting
3. create dimension > Save > Error
Selection_006

This PR add permission for allow create analytic tag dimension


pedrobaeza (approved):
I see this one correct due to the new v15 change in ir.model accesses, so merging it:

/ocabot merge patch

PR #579, opened 2023-07-23
[15.0][FIX] mrp_analytic: remove duplicate field

ps-tubtim:
- Add an Analytic Tags field to Manufacturing Order
- Remove an Analytic Account field because the field is already in core.

**Before**
Selection_119

**After**
Selection_120


AaronHForgeFlow (approved):
Removing the duplicate field and the dependency fixing should be done for sure. Ok with the new feature for the analytic tags, the improvement makes sense.

Saran440 (approved):
LGTM :+1:

PR #580, opened 2023-07-24
[15.0][ENH] stock_analytic: select analytic when scrapping product

ps-tubtim:
This PR is used to select Analytic Account and Analytic Tags when scrapping products.
Selection_123


AaronHForgeFlow (approved):
LGTM

PR #587, opened 2023-08-24
[16.0][ADD] account_analytic_tag_distribution: New module

victoralmau:
New module (related to https://github.com/OCA/account-analytic/pull/572#issuecomment-1688225050)

Locked by:
- [x] `account_analytic_tag`: https://github.com/OCA/account-analytic/pull/572

Please @pedrobaeza and  @chienandalu can you review it?

@Tecnativa TT44062


account_analytic_tag_distribution/models/account_analytic_tag.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2023 Tecnativa - Víctor Martínez
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import api, fields, models
+
+
+class AccountAnalyticTag(models.Model):
+    _name = "account.analytic.tag"
+    _inherit = ["account.analytic.tag", "analytic.mixin"]
```
pedrobaeza, line 8: Why the mixin here?

account_analytic_tag_distribution/models/account_move.py
```diff
@@ -0,0 +1,28 @@
...
+                for distribution in tags.mapped("analytic_distribution"):
+                    for d_key in list(distribution.keys()):
+                        if d_key not in item_distribution:
+                            item_distribution[d_key] = distribution[d_key]
+                        else:
+                            item_distribution[d_key] += distribution[d_key]
+                item.analytic_distribution = item_distribution
+
+    def action_post(self):
+        """Before post, set the analytical distribution according to analytic tags."""
```
pedrobaeza: Why before post?

account_analytic_tag_distribution/models/account_analytic_tag.py
victoralmau: The objective (as discussed in https://github.com/OCA/account-analytic/pull/572#issuecomment-1688225050) is to add the analytic distribution in tags, `analytic.mixin` has already defined the fields for it, avoiding adding code for the same purpose. Do you think it should be done with another way?

pedrobaeza: Which fields are you talking about?

victoralmau: The `analytic_distribution` field is defined by https://github.com/odoo/odoo/blob/16.0/addons/analytic/models/analytic_mixin.py, this field allows using the analytic_distribution widget to define the analytic distribution on the tag.

pedrobaeza: OK, I see. I was thinking on the traditional one2many field as in v15, but it's OK this way.

account_analytic_tag_distribution/models/account_move.py
victoralmau: In that method call `_create_analytic_lines()` method to create the analytic items, I think it is better to use `_create_analytic_lines()` directly, I will change it.

victoralmau:
Changes done.

account_analytic_tag_distribution/models/account_move_line.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2023 Tecnativa - Víctor Martínez
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    def _prepare_analytic_lines(self):
+        vals = super()._prepare_analytic_lines()
```
pedrobaeza: Here we have a conflict: what happens if there's a distribution in the standard analytic and also in the tag? I think we should disable and empty the standard analytic if we select a tag with distribution, and don't call super for that cases here.

account_analytic_tag_distribution/models/account_move_line.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    @api.onchange("analytic_tag_ids")
+    def _onchange_analytic_tag_ids(self):
+        """If any tag has a distribution, we will disable the standard analytical
+        distribution."""
+        if any(tag.active_analytic_distribution for tag in self.analytic_tag_ids):
```
pedrobaeza:
```suggestion
        if any(self.analytic_tag_ids.mapped("active_analytic_distribution")):
```

account_analytic_tag_distribution/models/account_move_line.py
```diff
@@ -0,0 +1,34 @@
...
+            self.analytic_distribution = False
+
+    def _prepare_analytic_lines(self):
+        """If any tag has distribution, we will not call super() to avoid conflicts
+        with standard flow."""
+        tags_with_distribution = self.analytic_tag_ids.filtered(
```
pedrobaeza, line 20: Is this `self.ensure_one()` for doing this?

pedrobaeza (approved):

chienandalu (approved):

PR #598, opened 2023-10-11
[13.0][IMP] purchase_analytic: avoid overwrite context in views inherited

ThiagoMForgeFlow:
Change the method of writing the context in the field order_line to avoid overwriting.

@ForgeFlow

### DIFF
--- a/purchase_analytic/__manifest__.py
+++ b/purchase_analytic/__manifest__.py
@@ -7,7 +7,7 @@
     "author": "Acsone SA/NV, Odoo Community Association (OCA)",
     "category": "Purchase Management",
     "website": "https://github.com/OCA/account-analytic",
-    "depends": ["purchase"],
+    "depends": ["purchase", "base_view_inheritance_extension"],
     "data": ["views/purchase_views.xml"],
     "license": "AGPL-3",
     "installable": True,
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):
Just for curiosity, what is the other module that needs to put a context there? Is that the purchase standard app?

ThiagoMForgeFlow:
> Just for curiosity, what is the other module that needs to put a context there? Is that the purchase standard app?

I have found the case with the purchase_blanket_order module from purchase-workflow repository

LoisRForgeFlow (approved):
LGTM

This is probably a major change as it adds a new dependency.

AaronHForgeFlow:
Merging to prevent conflicts between oca modules

/ocabot merge major

AaronHForgeFlow:
I just realized that the purchase_blanket_order is overwriting the context same way as this one was doing. I think that module should be fixed as well to inherit from base_view_inheritance_extension. Is that correct?

ThiagoMForgeFlow:
> I just realized that the purchase_blanket_order is overwriting the context same way as this one was doing. I think that module should be fixed as well to inherit from base_view_inheritance_extension. Is that correct?

Correct, the PR is still open: https://github.com/OCA/purchase-workflow/pull/2031

PR #581, opened 2023-07-28
[16.0][mig] analytic_partner

aktiv-heli-kantawala:

Issue #489, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [x] account_analytic_parent - By @mohammadazeem786 - #501
- [ ] account_analytic_required - By @nguyenminhchien - #541
- [x] account_analytic_sequence - By @victoralmau - #629
- [ ] account_analytic_tag_default
- [x] account_move_update_analytic - By @hbrunn - #559
- [ ] analytic_activity_based_cost - By @Urvisha-OSI - #538
- [x] analytic_base_department - By @hbrunn - #658
- [ ] analytic_partner - By @dreispt - #609
- [ ] analytic_tag_dimension
- [ ] mrp_analytic - By @kjoeldevk - #494
- [x] patch - By @hbrunn - #559
- [x] pos_analytic_by_config - By @chienandalu - #610
- [ ] procurement_mto_analytic
- [x] product_analytic - By @CLaurelB - #594
- [x] product_analytic_sale - By @rousseldenis - #621
- [ ] purchase_analytic
- [x] stock_landed_costs_analytic - By @rov-adhoc - #618
- [ ] stock_picking_analytic - By @yostashiro - #527

Missing module? Check https://github.com/OCA/maintainer-tools/wiki/%5BFAQ%5D-Missing-modules-in-migration-issue-list

KM-joel:
Migration to version 16.0 
- mrp_analytic #494 
@sbidoul

mohammadazeem786:
Migration to version 16.0

- account_analytic_parent https://github.com/OCA/account-analytic/pull/501

nicomacr:
Migration to version 16.0

- stock_analytic #506 
- purchase_stock_analytic #509 
- sale_stock_analytic #511

RodrigoBM:
PR [[MIG] purchase_analytic: Migration to 16.0](https://github.com/OCA/account-analytic/pull/524)

Raul-S73:
@sbidoul Migration to version 16.0:
- mrp_analytic_sale_project https://github.com/OCA/account-analytic/pull/549

rafaelbn:
`account_analytic_required` not needed in Odoo 16 as is included in core. Explained in https://github.com/OCA/account-analytic/pull/541#pullrequestreview-1367483412

miquelalzanillas:
Working on `account_analytic_tag_default`

PR #609, opened 2023-11-18
[16.0][mig] analytic_partner

dreispt:
Supersedes https://github.com/OCA/account-analytic/pull/581


analytic_partner/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+#. Go to **Invoicing > Configuration > Management > Analytic Items**.
```
chienandalu, line 1:
```suggestion
#. Go to **Invoicing > Accounting > Management > Analytic Items**.
```

analytic_partner/models/account_move_line.py
```diff
@@ -8,7 +8,8 @@
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
-    def _prepare_analytic_line(self):
-        res = super()._prepare_analytic_line()
-        res[0]["other_partner_id"] = self.move_id.partner_id.commercial_partner_id.id
+    def _prepare_analytic_lines(self):
```
chienandalu, line 11: I think that `_prepare_analytic_distribution_line` is the proper method to hook into now.

chienandalu (commented):
HI @dreispt I'm struggling to find out why this module is still needed. Searching for previous internal issues the problem seemed to be that when the analytic account used had a partner set it took that one instead of the one in the move line...

But I can't reproduce that issue anymore. What case are we covering here right now? As far as I could test, the other partner is always the same as the partner.

analytic_partner/models/account_move_line.py
maisim: Hi @chienandalu , I think you are right, we can add `other_partner_id` to the dict returned by _prepare_analytic_distribution_line.
Somethings like:
```
def _prepare_analytic_distribution_line(self, distribution, account_id, distribution_on_each_plan):
    vals = super()._prepare_analytic_distribution_line(distribution, account_id, distribution_on_each_plan)
    vals['other_partner_id'] = self.move_id.partner_id.commercial_partner_id.id
    return vals
```
Hi @dreispt, How do you want to proceed? Can I create a PR on your fork?

maisim (changes requested):

flotho:
Hi @maisim , thanks for review with @chienandalu 
Ad @dreispt doesn't seems to have new commits since a long time, I think you could use his branch to create a new PR  to supersedes this one. Authoring will be kept anyway.
Regards

dreispt:
@maisim Yes, you are welcome to pull this branch into your fork and make your own PR.

maisim:
There is another PR (a third) to migrate this module to 16.0 : https://github.com/OCA/account-analytic/pull/672
It seems almost finish \o/

Can someone update the Migration to 16.0 issue https://github.com/OCA/account-analytic/issues/489 and close  the current PR https://github.com/OCA/account-analytic/pull/609  ?

PR #610, opened 2023-11-21
[16.0][MIG] pos_analytic_by_config: Migration to 16.0

chienandalu:
WIP: Refactor to adapt to the new analytic distributions approach wich would lead to simpler code:

- New pos_config_id criteria in the analytic distribution model so we can add that weight.
- We can prefix the account so only sales go to that distribution and use a miscellaneous applicability domain in the plan
- With those two changes, we can already set the distributions in our PoS sales.

TODO:

- [x] Check invoicing from process
- [x] Redact readme
- [x] Refactor tests
- [x] Migration scripts

please check @pedrobaeza @victoralmau 

@Tecnativa TT45177


chienandalu:
Tests are failing on  `product_analytic` which I think isn't needed having the analytic distribution applicability criterias for product and product category.

I added a commit to ensure the compatibility of both of them

chienandalu:
Ready for review

apineux (approved):

FrancoMaxime (approved):
LGTM: code review & functional review

I agree with this new way of doing this, i started yesterday to continue your pr, but instead of using _generate_pos_order_invoice to inject the context i used _create_invoice.

Thanks for everything

chienandalu:
> I agree with this new way of doing this, i started yesterday to continue your pr, but instead of using _generate_pos_order_invoice to inject the context i used _create_invoice.

Yes, `_generate_pos_order_invoice` is also involved in the generation of other associated moves that (I'm not really not sure about it) could be interesting to filter in a distribution for someone

pedrobaeza:
@chienandalu it seems the conversion from MD to RST didn't go well. Please check the texts.

chienandalu:
Let's try again

pedrobaeza:
Still the same... Not sure which is the problem. Maybe you should keep it as RST.

chienandalu:
Let's try again

PR #612, opened 2023-12-07
[12.0][ADD] account_analytic_root

victor-champonnois:
Add a top parent for analytic account, for reporting purposes. 
ported to 16 : https://github.com/OCA/account-analytic/pull/622

@robinkeunen


account_analytic_top_parent/models/account_analytic_account.py
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: 2022 Coop IT Easy SC
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import api, fields, models
+
+
+class AccountAnalytic(models.Model):
+    _inherit = 'account.analytic.account'
```
robinkeunen: It looks like you did not run pre-commit locally.

account_analytic_top_parent/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,29 @@
...
+- Account A / Account AB / Account ABC
+- Account B
+- Account C / Account CB
+
+If we group by parent analytic account we get:
+
+- Account A 
+- Account A / Account AB 
+- Account C / Account CB
+- no value / Account B
```
robinkeunen: Wouldn't it be :

- Account A
- Account A / Account AB
- False
- Account C

robinkeunen (changes requested):
Great description and clean code. I think there is an error in the example though.

robinkeunen:
Also, a unit test would be great :-)

account_analytic_top_parent/readme/DESCRIPTION.rst
victor-champonnois: yes correct

victor-champonnois:
@robinkeunen unit test added and description updated.

robinkeunen:
@victor-champonnois I think @carmenbianca could help with the CI. Still, it looks like you did not run precommit locally.

account_analytic_top_parent/tests/test_account_invoice.py
```diff
@@ -0,0 +1,32 @@
...
+            )
+        self.analytic_account_ABC = self.env['account.analytic.account'].create(
+            {'name': 'test analytic_account_AB',
+             'parent_id': self.analytic_account_AB.id,}
+            )
+
+
+    def test_top_parent_if_no_parent_returns_itself(self):
+        self.assertEqual(self.analytic_account_A.top_parent_id, 
+                         self.analytic_account_A.top_parent_id)
```
robinkeunen: you're testing X = X, don't you mean :
```suggestion
        self.assertEqual(self.analytic_account_A.top_parent_id, 
                         self.analytic_account_A)
```

robinkeunen (changes requested):
I think there is a mistake in the tests.

+ run precommit

victor-champonnois:
@robinkeunen fixed the test (as in https://github.com/OCA/account-analytic/pull/622), but not precommit (it's not setup in this repo)

robinkeunen (approved):
You can rebase on https://github.com/OCA/account-analytic/pull/624 once merged. I'll approve this one in the meantime. Thx!

AaronHForgeFlow:
Hi, just FYI, in the module project_wbs there is a field "Root analytic account" that is same functionality:

https://github.com/OCA/project/blob/12.0/project_wbs/models/account_analytic_account.py#L158

I don't mind to adapt the project_wbs module to be dependent on this one from v16 onwards. But for v12 I won't be doing it. 

Any comments would be appreciated.

robinkeunen:
I did not know about WBS, thanks for the pointer. Can these two modules live in the same repo or should @victor-champonnois mark them as rebel ?

Also, is the name `top_parent_id` good enough to be used in `project_wbs` ?

AaronHForgeFlow:
@robinkeunen I think adding it as rebel module is not needed, as long the field names are different. 

I guess in v16 I will need to add a migration script to rename the field from 'project_analytic_id' to 'top_parent_id'.

robinkeunen:
I mean it's still time to call it `root_analytic_account_id` if that's more consistent with your use case. It makes to meaningful difference on our side.

victor-champonnois:
@AaronHForgeFlow Thank you for the info Aaron. 
To me there is a limitation in the _project_analytic_id_ field. _project_analytic_id_ from _project_wbs_ computes the root analytic account for a project. So the information from account analytic line to _project_analytic_id_ is through the _project_id_ on the line. So it only works for timesheets (only account analytic lines that are timesheets have a project defined). Analytic lines that are not timesheets (created from account move lines) cannot have access to this root analytic account. 

I would say that this current implementation is a bit more generic, since it rely on analytic account, which are defined on all analytic lines. 

To me it's OK to drop the current PR if it's in conflicts with project_wbs. However regarding the port to 16, I think project_wbs could be refactored in 16 to depends on this module. We can rename the field and the module for a better name if needed. "root_analytic_account" is indeed a better name.  
Also, having the functionality in the _account-analytic_ repo rather than in _project_ makes it more discoverable.

AaronHForgeFlow:
@victor-champonnois , Thanks for the explanation. It is what I thought. I will do the refactor in the project_wbs module in v16. I like the name the field "Root Analytic Account". You can use that name and then I can make a migration script in project_wbs module to use the new name. 

For versions < 16 I will not propose any change, at the end, as you said, there is no actual overlapping, as the field in the project_wbs module is only for analytic accounts linked to a project. It is ok to me you can continue with this PR.

robinkeunen:
@victor-champonnois let's go for `root_analytic_account_id` then, it will save us a migration script between 12 and 16.

victor-champonnois:
@AaronHForgeFlow @robinkeunen I renamed the module `account_analytic_root` and the field `root_id` (in order to be consistent with `parent_id`). 
I will do the same thing on the 16.0 PR.

robinkeunen (approved):

robinkeunen:
Can you rebase on 12.0 ?

victor-champonnois:
rebased and renamed the commit

dreispt (approved):

dreispt:
Merged with #657

PR #613, opened 2023-12-07
[12.0][IMP] Add config setting to always use product analytic

victor-champonnois:
Add a configuration setting to always use the product analytic account and overide the default (for instance the sale order analytic account).

This is useful when we want to do sales reporting with analytic account based on product, instead of Sale orders. 

Ported to 16  https://github.com/OCA/account-analytic/pull/623
@robinkeunen

### DIFF
--- a/product_analytic/models/account_invoice.py
+++ b/product_analytic/models/account_invoice.py
@@ -12,6 +12,7 @@ INV_TYPE_MAP = {
     'in_refund': 'expense',
 }
 
+GROUP_AUPAA = 'product_analytic.group_always_use_product_analytic_account'
 
 class AccountInvoiceLine(models.Model):
     _inherit = 'account.invoice.line'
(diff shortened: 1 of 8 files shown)


product_analytic/models/account_invoice.py
```diff
@@ -31,8 +32,8 @@ def _onchange_product_id(self):
     def create(self, vals_list):
         for vals in vals_list:
             inv_type = self.env.context.get('inv_type', 'out_invoice')
-            if vals.get('product_id') and inv_type and \
-                    not vals.get('account_analytic_id'):
+            replace_account_analytic = self.env.user.has_group(GROUP_AUPAA) or not vals.get('account_analytic_id')
+            if vals.get('product_id') and inv_type and replace_account_analytic:
```
robinkeunen, line 36: I had trouble understanding this bit. I think unit tests would help.

robinkeunen (changes requested):
The code seems correct but a few unit tests would help.

victor-champonnois:
@robinkeunen unit tests added

victor-champonnois:
ported to 16 in https://github.com/OCA/account-analytic/pull/623

PR #616, opened 2024-01-02
[16.0][MIG]sale_stock_analytic

moitabenfdz:


sale_stock_analytic/models/stock_rule.py
```diff
@@ -0,0 +1,39 @@
...
+            location_id,
+            name,
+            origin,
+            company_id,
+            values,
+        )
+        sol_id = move_values.get("sale_line_id", False)
+        if sol_id:
+            sol_model = self.env["sale.order.line"]
+            sol = sol_model.browse(sol_id)
```
cvinh, line 35: We should consider the case where the analytic account is at the sale order level. In that case, we have to take the sale order analytic account on every line of the stock move

cvinh (changes requested):

sale_stock_analytic/models/stock_rule.py
cvinh: I did a PR on your branch
https://github.com/DynAppsNV/account-analytic/pull/2

AungKoKoLin1997: I don't think we have to do this. Passing `analytic_distribution` to stock move is enough.

sale_stock_analytic/tests/test_sale_stock_analytic.py
```diff
@@ -0,0 +1,42 @@
...
+
+class TestSaleStockAnalytic(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.sale_order_model = cls.env["sale.order"]
+        cls.sale_order_line_model = cls.env["sale.order.line"]
+        cls.product_model = cls.env["product.product"]
+        cls.res_partner_model = cls.env["res.partner"]
+
```
sergiobstoj, line 15: Try to avoid whitespaces in-between methods

PR #614, opened 2023-12-21
[17.0][MIG] stock analytic: Migration to 17.0

moitabenfdz:

PR #617, opened 2024-01-02
[17.0]]MIG]sale_stock_analytic: Migration to 17.0

moitabenfdz:
Depends on:

- [ ] #614


sale_stock_analytic/models/stock_rule.py
```diff
@@ -0,0 +1,38 @@
...
+            company_id,
+            values,
+        )
+        sol_id = move_values.get("sale_line_id", False)
+        if sol_id:
+            sol_model = self.env["sale.order.line"]
+            sol = sol_model.browse(sol_id)
+            analytic_distribution = sol.analytic_distribution
+            if analytic_distribution:
+                move_values.update({"analytic_distribution": analytic_distribution})
```
cvinh, line 37: You might take into account the case where analytic account is at the sale order level...
you might cherry-pick https://github.com/DynAppsNV/account-analytic/pull/2

cvinh (changes requested):

sale_stock_analytic/models/stock_rule.py
aisopuro: Alternatively, Odoo's own logic merges the two when preparing the invoice line information from a sale line: https://github.com/odoo/odoo/blob/1648ee1abff1ae12b3933baa85a2e048fa2eb557/addons/sale/models/sale_order_line.py#L1142C9-L1142C35.

PR #618, opened 2024-01-04
[16.0][MIG] stock_landed_costs_analytic: Migration to 16.0

rov-adhoc:


jjscarafia (approved):
LGTM

stock_landed_costs_analytic/__manifest__.py
```diff
@@ -7,11 +7,12 @@
         This module adds an analytic account and analytic tags on landed costs
         lines so that on landed costs validation account moves get analytic
         account and analytic tags values from landed costs lines.""",
-    "version": "13.0.1.0.0",
+    "version": "16.0.1.0.0",
     "license": "AGPL-3",
     "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/account-analytic",
-    "depends": ["stock_landed_costs"],
+    "depends": ["stock_landed_costs", "account_analytic_tag"],
```
pedrobaeza: The module should work without this extra module with the base analytic that Odoo uses. Then, you can put an extra module for handling tags.

pedrobaeza (changes requested):

stock_landed_costs_analytic/__manifest__.py
rov-adhoc: @pedrobaeza done

pedrobaeza (approved):
/ocabot merge nobump
/ocabot migration stock_landed_costs_analytic

PR #622, opened 2024-01-29
[16.0][MIG] account_analytic_root

victor-champonnois:
Back-ported to 12 in https://github.com/OCA/account-analytic/pull/612
Add a top parent for analytic account, for reporting purposes.


victor-champonnois:
@robinkeunen

robinkeunen (changes requested):
Migration is ok but I have the same reserves as in #612 

By labelling this PR [ADD] and label #612 as a backport, you might get more attention.

victor-champonnois:
@robinkeunen changed the PR name as you suggested and fixed the test.

robinkeunen (approved):

victor-champonnois:
renamed to `account_analytic_root` (see discussion https://github.com/OCA/account-analytic/pull/612)

AaronHForgeFlow:
Sorry for the nitpicking @victor-champonnois but the related model should be in the name of the field, as it is a Many2one. It should be root_analytic_account_id.

victor-champonnois:
@AaronHForgeFlow Changed the name as you suggested.

AaronHForgeFlow (approved):
LGTM! What about displaying the root analytic account in the form view?

PR #623, opened 2024-01-31
[16.0][IMP] Add config setting to always use product analytic

victor-champonnois:
This module adds an option to force the use of the product analytic account for service products invoiced on timesheets. 

Use case: 
We have a service product, invoiced on timesheet and that triggers the creation of a project when a SO is confirmed. Note that doing this, an analytic account is created and linked to the SO (module _sale_project_)
An income analytic account is defined on this product. We want this analytic account to be in the analytic distribution of the invoice lines that are created from the SO. 
Now, this is not possible. When creating an invoice, the analytic distribution on the invoice lines is the SO analytic account.   
This option forces the use of the product analytic account on the invoice line instead. 

I am not sure if this is the right place to make this change, since this module _product_analytic_ does not depend on _sale_project_. Tell me if you think of a better place for this code.   

Forward porting of https://github.com/OCA/account-analytic/pull/613


victor-champonnois:
@robinkeunen

robinkeunen (approved):
Code looks good and thx for the unit test.

I think it could be useful to document the problematic use case (documented in the PR description) in the commit message.

victor-champonnois:
@robinkeunen use case added in the commit message

PR #634, opened 2024-03-13
[17.0][MIG] account_analytic_parent

RicardCForgeFlow:
Migration to 17.0

@ForgeFlow


account_analytic_parent/views/account_analytic_account_view.xml
```diff
@@ -0,0 +1,30 @@
...
+    -->
+    <record id="view_account_analytic_account_form" model="ir.ui.view">
+        <field name="name">account.analytic.account.form</field>
+        <field name="model">account.analytic.account</field>
+        <field name="type">form</field>
+        <field name="inherit_id" ref="analytic.view_account_analytic_account_form" />
+        <field name="arch" type="xml">
+            <field name="plan_id" position="after">
+                <field name="parent_id" string="Parent Analytic Account" />
+            </field>
```
GuillemCForgeFlow, line 14:
```suggestion
            <group name="main" position="attributes">
                <attribute name="string">General</attribute>
            </group>
            <group name="main" position="before">
                <group name="parent" col="2">
                    <field name="parent_id" class="oe_inline"/>
                </group>
            </group>
```
As a general improvement. I guess that it makes more sense to have the view in a similar structure as in the Product Category, where we also have a Parent-Child relation.
In order to adapt, we also could name the original group with the "General" string. What do you think?

GuillemCForgeFlow (commented):
Also, I believe it would be great to add a CONFIGURE.md file detailing that the user accessing to the Analytic Accounting features needs to have the `Technical / Analytic Accounting` group added.
The rest, code review LGTM 👍🏽

account_analytic_parent/static/description/index.html
```diff
@@ -0,0 +1,448 @@
+<?xml version="1.0" encoding="utf-8"?>
```
AndreuOForgeFlow, line 1:
```suggestion

```
XML declarations in HTML module descriptions are deprecated in Odoo 17

AndreuOForgeFlow (approved):
Very small suggestion. Other than that LGTM! It would be nice to add in the README file the path to Analytic Accounting: Accounting > Configuration > Analytic Accounting etc.

dreispt (approved):

PR #637, opened 2024-03-26
[16.0][MIG] procurement_mto_analytic: Migration to 16.0

AungKoKoLin1997:
@qrtl QT4346


procurement_mto_analytic/models/stock_move.py
```diff
@@ -0,0 +1,18 @@
...
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _prepare_procurement_values(self):
+        res = super()._prepare_procurement_values()
+        analytic_distribution = self.sale_line_id.analytic_distribution
+        # In case of using stock_analytic and mrp_stock_analytic
+        if not analytic_distribution and hasattr(self, "analytic_distribution"):
+            analytic_distribution = self.analytic_distribution
```
yostashiro:
```suggestion
```
Let's remove this as `stock_analytic` already takes care of this: https://github.com/OCA/account-analytic/blob/14ecb3faf4d6ee0e41fe47bcbe1a8a3d9883d085/stock_analytic/models/stock_move.py#L34-L46

procurement_mto_analytic/models/stock_rule.py
```diff
@@ -0,0 +1,39 @@
...
+            matching_po_line = self._find_matching_analytic_distribution_record(
+                values["analytic_distribution"]
+            )
+            if matching_po_line:
+                domain += (("order_line", "in", tuple(matching_po_line)),)
+            else:
+                # To create new PO
+                domain += (("order_line", "=", 0000000),)
+        else:
+            domain += (("order_line.analytic_distribution", "=", False),)
```
yostashiro: Can you please check if this works.

```suggestion
    def _make_po_get_domain(self, company_id, values, partner):
        domain = super()._make_po_get_domain(company_id, values, partner)
        matching_lines = self.env["purchase.order"].sudo().search([dom for dom in domain]).order_line.filtered(lambda x: x.analytic_distribution == values.get("analytic_distribution"))
        if matching_lines:
            domain += (("order_line", "in", matching_lines.ids),)
        else:
            # To create new PO
            domain += (("order_line", "=", 0000000),)
```

yostashiro (changes requested):

procurement_mto_analytic/models/stock_rule.py
```diff
@@ -0,0 +1,27 @@
...
+                lambda x: x.analytic_distribution == values.get("analytic_distribution")
+            )
+        )
+        if matching_lines:
+            domain += (("order_line", "in", tuple(matching_lines.ids)),)
+        else:
+            domain += (
+                ("order_line", "=", False),
+            )  # This might need adjustment based on your intent.
+        return domain
```
yostashiro: Reviewing the design, I now doubt that this should be the way to go, as there can be cases where one purchase order holds mutplie analytic distributions. We should not be extending this method.

We should instead create a PR in odoo/odoo to split https://github.com/OCA/OCB/blob/1e9b8db/addons/purchase_stock/models/stock_rule.py#L125 into a method, and extend that method in this module to add analytic distribution as a key.

yostashiro (changes requested):

procurement_mto_analytic/models/stock_rule.py
```diff
@@ -6,22 +7,25 @@
...
-            )  # This might need adjustment based on your intent.
-        return domain
+    def _get_po_line_grouping_key(self, line):
+        """
+        Extend the grouping logic to conditionally include the analytic_distribution
+        only if it exists.
+        """
+        original_key = super()._get_po_line_grouping_key(line)
+        if line.analytic_distribution:
+            analytic_distribution_key = tuple(line.analytic_distribution.items())
```
aisopuro: In Python, dict entries are (basically) random-ordered in versions 3.6 and below, and insertion-ordered after that. So we have the following:
```
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.16.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: a = {"1": 100, "2": 100}
In [2]: b = {"2": 100, "1": 100}
In [3]: tuple(a.items()) == tuple(b.items())

Out[3]: False

In [4]: a.items()
Out[4]: dict_items([('1', 100), ('2', 100)])

In [5]: b.items()
Out[5]: dict_items([('2', 100), ('1', 100)])
```
To me, this intuitively doesn't make sense: if I have a distribution with the same accounts and the same percentages, those should be the same regardless of the order, right? I suggest adding a `sorted` call into the key generation:

```
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.16.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: a = {"1": 100, "2": 100}
In [2]: b = {"2": 100, "1": 100}
In [3]: tuple(sorted(a.items())) == tuple(sorted(b.items()))

Out[3]: True
```

AungKoKoLin1997: > if I have a distribution with the same accounts and the same percentages, those should be the same regardless of the order, right? I suggest adding a sorted call into the key generation

@aisopuro items of `analytic_distribution` are always sorted by default. So, we don't need to worry about it.

aisopuro: Could you link to the place in Odoo's source code that ensures they are always sorted, regardless of whether:
* The distribution was added by a user through the UI
* The distribution was added by business logic in Python
* The distribution was imported via excel
* etc

Or does the `JSONB` column type in Postgresql guarantee keys are ordered?

AungKoKoLin1997: Depends on this source(https://www.postgresql.org/docs/10/datatype-json.html#JSON-TYPE-MAPPING-TABLE:~:text=There%20are%20two,value%20is%20kept.), I conclude analytic_distribution are sorted depends on the efficiency.
Whenever we insert the key in any order, it will be sorted if they have same data.
Eg, when we insert {"1":60, "2":40} and {"2":40, "1":60}, they will be sorted. So, the result will be the same.
Screenshot 2024-05-17 at 10 23 02

aisopuro: Nice :+1:, my suggestion is then not needed.

kanda999 (approved):
Functional review.

yostashiro (approved):
Code review. LGTM.

aisopuro (commented):
Pre-commit checks are failing @AungKoKoLin1997 

I'm not a maintainer of this repo, so cannot merge this PR. 

Reviewing the changes since I last looked at this (https://github.com/OCA/account-analytic/compare/723dbf7..f032887e4fb06d34c0d5d4a0bf3a051887034736), the changes to `procurement_mto_analytic` seem fine.

AungKoKoLin1997:
> Pre-commit checks are failing @AungKoKoLin1997

@aisopuro This is not related with my PR. I think dotfiles for 16.0 should be updated.

```
Generate default setup.py for an addons directory........................Failed
- hook id: setuptools-odoo-make-default
- exit code: 1

Traceback (most recent call last):
  File "/home/runner/.cache/pre-commit/repobcocfbyg/py_env-python3/bin/setuptools-odoo-make-default", line 3, in <module>
    from setuptools_odoo.make_default_setup import main
  File "/home/runner/.cache/pre-commit/repobcocfbyg/py_env-python3/lib/python3.11/site-packages/setuptools_odoo/__init__.py", line 6, in <module>
    from .core import get_addon_metadata
  File "/home/runner/.cache/pre-commit/repobcocfbyg/py_env-python3/lib/python3.11/site-packages/setuptools_odoo/core.py", line 15, in <module>
    from . import base_addons, external_dependencies
  File "/home/runner/.cache/pre-commit/repobcocfbyg/py_env-python3/lib/python3.11/site-packages/setuptools_odoo/base_addons.py", line 8, in <module>
    from pkg_resources import resource_string
ModuleNotFoundError: No module named 'pkg_resources'

Generate requirements.txt for an addons directory........................Failed
- hook id: setuptools-odoo-get-requirements
- exit code: 1

Traceback (most recent call last):
  File "/home/runner/.cache/pre-commit/repobcocfbyg/py_env-python3/bin/setuptools-odoo-get-requirements", line 3, in <module>
    from setuptools_odoo.get_requirements import main
  File "/home/runner/.cache/pre-commit/repobcocfbyg/py_env-python3/lib/python3.11/site-packages/setuptools_odoo/__init__.py", line 6, in <module>
    from .core import get_addon_metadata
  File "/home/runner/.cache/pre-commit/repobcocfbyg/py_env-python3/lib/python3.11/site-packages/setuptools_odoo/core.py", line 15, in <module>
    from . import base_addons, external_dependencies
  File "/home/runner/.cache/pre-commit/repobcocfbyg/py_env-python3/lib/python3.11/site-packages/setuptools_odoo/base_addons.py", line 8, in <module>
    from pkg_resources import resource_string
ModuleNotFoundError: No module named 'pkg_resources'
```

aisopuro:
OK. But again, I'm not a maintainer of this repo, so you should probably
communicate this to an actual maintainer.

AungKoKoLin1997:
> I think dotfiles for 16.0 should be updated.

I create a PR to update dotfiles.
https://github.com/OCA/account-analytic/pull/882

PR #641, opened 2024-04-10
[16.0][IMP] stock_analytic: salvage analytic distribution from procurement

AungKoKoLin1997:
@qrtl QT4346

### DIFF
--- a/stock_analytic/models/__init__.py
+++ b/stock_analytic/models/__init__.py
@@ -3,4 +3,5 @@
 from . import analytic_applicability
 from . import stock_move
 from . import stock_picking
+from . import stock_rule
 from . import stock_scrap
--- a/stock_analytic/models/stock_rule.py
+++ b/stock_analytic/models/stock_rule.py
@@ -1,0 +1,11 @@
+# Copyright 2024 Atte Isopuro <atte.isopuro@avoin.systems>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class StockRule(models.Model):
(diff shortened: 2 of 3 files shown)


rousseldenis (approved):
Code review.

yostashiro (approved):
Code review.

StefanRijnhart:
Is it perhaps the case that this is the same issue that is solved in the 17.0 migration using `_get_custom_move_fields`? See https://github.com/OCA/account-analytic/pull/614/commits/fa50c404e9a89e43bb5ea1c65a9ff1166180cde5, included in https://github.com/OCA/account-analytic/pull/614). If so, isn't using that entry point more appropriate?

yostashiro:
> Is it perhaps the case that this is the same issue that is solved in the 17.0 migration using `_get_custom_move_fields`? See [fa50c40](https://github.com/OCA/account-analytic/commit/fa50c404e9a89e43bb5ea1c65a9ff1166180cde5), included in #614). If so, isn't using that entry point more appropriate?

Thanks! The suggested approach does look better indeed. @AungKoKoLin1997

StefanRijnhart:
@AungKoKoLin1997  thanks for the update! Can you credit @aisopuro as a co-author of the commit? (see https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line)

StefanRijnhart (approved):
Thanks for the follow-up!

AaronHForgeFlow (approved):

PR #644, opened 2024-04-16
[17.0][MIG] mrp_stock_analytic

aisopuro:
## Depends on
* #614
* #682

## Notes
* Includes an additional `[IMP]` commit since it wasn't strictly related to the migration. It can be squashed if the maintainers feel it is better that way


aisopuro:
Rebased on top of 17.0. @StefanRijnhart I think this is still active, are you able to mark it not stale?

StefanRijnhart:
@aisopuro We should be fine if we keep on being active on it! Can you check the failing tests?

aisopuro:
I took a look, but the trace suggests the failure is in `stock_analytic`, which is green in `17.0`. So its probably some kind of non-trivial interaction, so I unfortunately probably won't manage to get to it anytime soon.

StefanRijnhart:
@aisopuro then this one will become stale again I'm afraid

mrp_stock_analytic/tests/test_mrp_stock_analytic.py
```diff
@@ -0,0 +1,148 @@
+# Copyright 2023 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import Form
+
+from odoo.addons.stock_analytic.tests.test_stock_picking import TestStockPicking
+
+
+class TestMrpStockAnalytic(TestStockPicking):
```
rousseldenis: @aisopuro This is not correct.

As the imported class will launch again tests from stock_analytic module.

Either you don't depend on it (TransactionCase instead), either you should first do a PR (we can quickly merge it) to do a common class for the setup in stock_analytic module.

StefanRijnhart: Good find! It *can* be a valid approach to ensure that the original tests still work after installation of this module. Is that what was intended @yostashiro?
If not, then maybe the extraction of a common setup class can still be included in this PR (if in a separate commit)?

yostashiro: > Good find! It can be a valid approach to ensure that the original tests still work after installation of this module. Is that what was intended @yostashiro?

No, I wasn't intending that. I just thought I'd use something available, which was the wrong approach I guess. I apologise for the trouble.

aisopuro: Thanks for the help @rousseldenis . Opened #682 and marked this as depending on it

aisopuro: Oh, seems it didn't fix it :disappointed: . Looks like this only shows up when running the at-install tests in a fresh database. I don't get the error on a second run in an already installed database.

rousseldenis: In fact, this is partly better. The error is no more launched by mrp_stock_analytic tests but in stock_analytic only

StefanRijnhart: @yostashiro Thanks for clarifying!

aisopuro: I think I've tracked down the cause of the issue, but I'm not sure how to fix/avoid it.

Symptoms:
* When I install a fresh database and install `stock_analytic`, its tests pass
* When I install a fresh database and install `mrp_stock_analytic` and run the tests for `stock_analytic` its tests fail

Based on debugging, the issue seems to be that `mrp_stock_analytic` depends on both `mrp` and `stock` (as it should), and that makes Odoo load them in the order
1. `stock`
2. `mrp`
3. `stock_analytic`
4. `mrp_stock_analytic`

When `mrp` is loaded, it contains demo data that sets the default resupply on the test warehouse to happen through manufacture rather than shipping in new product through pickings: https://github.com/odoo/odoo/blob/a2909105db3e016bfbbd092c9c10498dc610cfc1/addons/mrp/data/mrp_data.xml#L18

This then replaces/deactivates the routes in such a way that the normal stock operations no longer work. I've tried a couple things (enabling the manufacturing route on the product, disabling the option the `mrp` module activates), but I either get the same error, or the analytics aren't propagated when the product is manufactured, because of course not, that is the job of `mrp_stock_analytic`, which is not yet loaded.

Any ideas for how to solve this issue would be appreciated.

aisopuro: Solved this, those interested can check https://github.com/OCA/account-analytic/pull/644/commits/cb3995423905eb12c1a652d314133c7791e5e235 for details.

aisopuro:
@StefanRijnhart I've fixed comments and rebased, this PR is good to go again.

mrp_stock_analytic/models/mrp_production.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2021 ACSONE SA/NV
+# Copyright 2023 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class MrpProduction(models.Model):
+    _name = "mrp.production"
+    _inherit = ["mrp.production", "analytic.mixin"]
```
StefanRijnhart: This seems redundant, as it is present in Odoo 17 because of this change: https://github.com/odoo/odoo/commit/f13d537a42ff1a2713484d0e7c39b73fbb87a7fb#diff-35844959b2eb04809756ac5ad61ac87e18b9064ee8ad4036b96e908adc598f88R11

mrp_stock_analytic/views/mrp_production_views.xml
```diff
@@ -0,0 +1,24 @@
...
+        </field>
+    </record>
+    <record id="mrp_production_tree_view" model="ir.ui.view">
+        <field name="name">mrp.production.tree</field>
+        <field name="model">mrp.production</field>
+        <field name="inherit_id" ref="mrp.mrp_production_tree_view" />
+        <field name="arch" type="xml">
+            <field name="analytic_distribution" position="attributes">
+                <attribute name="readonly">state in ('done', 'cancel')</attribute>
+                <attribute name="optional">show</attribute>
```
StefanRijnhart: Is there a compelling reason to override upstream's optional="hide"? It is also linked to the analytic accounting user group there. https://github.com/odoo/odoo/commit/f13d537a42ff1a2713484d0e7c39b73fbb87a7fb#diff-cd75111aeb0015ab66e8b516edbfc92044f852718ec73251f3fdbfee1b69cfbeR11

StefanRijnhart:
@aisopuro Great work on the tests, thanks!

mrp_stock_analytic/models/mrp_production.py
aisopuro: Seems to have been a leftover from the 16 version, at which point this maybe wasn't there. Changed in https://github.com/OCA/account-analytic/pull/644/commits/f18ed0b377c7cd48375b17a6546aa62174dbe3df.

mrp_stock_analytic/views/mrp_production_views.xml
aisopuro: Can't see one. It was originally added with the 16 version of the module, so maybe the field wasn't added yet at that point? I removed our override in https://github.com/OCA/account-analytic/pull/644/commits/d2f3771f2e00b4b77ba792034e26cb3465fe980f.

StefanRijnhart: Yes indeed, there was no analytic distribution in Odoo 16's mrp.production model. As such, the last two commits (maybe other commits as well?) can be squashed into the migration commit (and I will try very hard to make this my last comment :grin:)

aisopuro: Alright, squashed those commits into the migration commit, and the final test-fix into the earlier commit that added the test. I feel like the current set of commits are now separate changes, and I personally wouldn't squash them any further.

StefanRijnhart (approved):
Thanks for all the follow-up!

mrp_stock_analytic/views/mrp_production_views.xml
StefanRijnhart: Thank you, and agreed!

StefanRijnhart:
@rousseldenis Is this one ok for you now?

rousseldenis:
@aisopuro Just improve commit message to make module name appears:

image

aisopuro:
Renamed and rebased @rousseldenis

rousseldenis (approved):

PR #645, opened 2024-04-16
[17.0][FIX] stock_analytic: actually propagate distribution

aisopuro:
## Depends on
* #614


aisopuro:
Cherry-picked into #614, so I think this one is no longer needed.

PR #656, opened 2024-05-17
[14.0][FIX] stock_analytic: Do not use existing product categories, create new ones.

victoralmau:
Do not use existing product categories, create new ones.

Using existing data and modifying it may cause side effects in some cases, for example if there is already linked data.

@Tecnativa


rousseldenis (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #660, opened 2024-05-31
[IMP] stock_picking_analytic: improve the performance of the analytic distribution computation

CLaurelB:
Replaced the stock picking analytic distribution search and validations
with a query to improve performance when there are large amounts of
picking and move records. Also, updated test to handle a case where
there are different analytic accounts on moves related to a picking.

### DIFF
--- a/stock_picking_analytic/models/stock_picking.py
+++ b/stock_picking_analytic/models/stock_picking.py
@@ -26,9 +26,6 @@ class StockPicking(models.Model):
     # pylint: disable=method-search
     move_ids_without_package = fields.One2many(search="[]")
 
-    # We use move_ids_without_package instead of move_ids throughout this method to
-    # guarantee consistent behavior. i.e. the value of move_ids is available only after
-    # the record has been saved.
     @api.depends(
         "move_ids_without_package.analytic_distribution",
         "original_analytic_distribution",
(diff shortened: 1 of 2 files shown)


CLaurelB:
@luisg123v could you review, please?

stock_picking_analytic/models/stock_picking.py
```diff
@@ -37,18 +34,40 @@ def _compute_analytic_distribution(self):
         """
         Get analytic account from first move and put it on picking
         """
+        picking_model = self.env[self._name]
+        picking_model.flush_model(["move_ids", "original_analytic_distribution"])
```
luisg123v: Don't need the model, just use `self`

stock_picking_analytic/models/stock_picking.py
```diff
@@ -37,18 +34,40 @@ def _compute_analytic_distribution(self):
...
+                stock_picking AS picking
+            LEFT JOIN
+                stock_move AS move
+                ON picking.id = move.picking_id
+            WHERE
+                picking.id IN %s
+            GROUP BY
+                picking.id;
+            """,
+            (tuple(self.ids or [0]),),
```
luisg123v: When an empty list is passed, it could cause performance issues.

Just early return if `not self.filtered("id")`

On the other hand, pro tip:

You may pass params as a list to avoid having to provide a extra comman when tere's only one param.

Instead of:

`(param1,)`

It could be:

`[param1]`

luisg123v (changes requested):

stock_picking_analytic/models/stock_picking.py
```diff
@@ -37,18 +34,40 @@ def _compute_analytic_distribution(self):
         """
         Get analytic account from first move and put it on picking
         """
+        if not self.filtered("id"):
+            return
+        self.flush_model(["move_ids", "original_analytic_distribution"])
+        self.move_ids.flush_model(["analytic_distribution"])
```
luisg123v: This will trigger a read on the field.

luisg123v (changes requested):

stock_picking_analytic/models/stock_picking.py
luisg123v: :laughing: it's still reading the field.

Just after you access `self.move_ids`, it's reading the field.

CLaurelB: Third time's a charm

luisg123v (approved):
LGTM :+1: 

Hi @moylop260,

What do you think about this optimization?

This is to avoid a `MemoryError` while computing a field.

CC @desdelinux

moylop260:
Do you have stats about before/after this fix?

Is needed to use `apply_rules` and `active=True` filter?
e.g. https://github.com/odoo/odoo/pull/165010/files

moylop260:
In general LGTM since that you only need one record from the whole recordset

so psql is better to manage this

CLaurelB:
Hi @moylop260 

> Do you have stats about before/after this fix?

We do not have stats because before the change, the field computation for all records did not even finish and the process would die.

> Is needed to use `apply_rules` and `active=True` filter? e.g. https://github.com/odoo/odoo/pull/165010/files

Stored computed fields are already calculated as sudo, so rules and permissions do not apply. None of these models have an active field, so the active test also does not apply.

moylop260 (approved):
👍

PR #661, opened 2024-05-31
[16.0][FIX] pos_analytic_by_config: migration scripts

chienandalu:
cc @Tecnativa TT46020


pedrobaeza:
This should be prefixed with `[OU-ADD]`.

pedrobaeza (approved):

PR #663, opened 2024-06-12
[17.0][MIG] account_analytic_sequence: Migration to 17.0

BT-dherreros:


account_analytic_sequence/i18n/account_analytic_sequence.pot
```diff
@@ -0,0 +1,29 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* account_analytic_sequence
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 16.0\n"
```
BT-cserra, line 7: I see in the PO files the version is wrong (it's 10.0 in most of them). I don't know if we have to keep them because they indicate the version in which the translations were first added. However in the POT, it should be changed to the current version: 17.0.

BT-cserra (changes requested):
I'm not sure we still need the POT file. But if we need it, or if we provide it, then the version should be correct.

BT-anieto:
@BT-dherreros Missing https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

Although the precommit hook fails when you do not set the version 17 in the manifest, the change of the version should appear in the [MIG] commit and not in the [UPD]

CasVissers-360ERP (approved):
Functional review

gurneyalex (approved):

PR #671, opened 2024-07-08
[17.0][MIG] purchase_request_analytic: Migration to 17.0

StefanRijnhart:
For lack of a 16.0 version, the migration from analytic_account_id to analytic_distribution is done here.


rousseldenis:
@StefanRijnhart Thanks. Maybe changes could have been put in separate commit in order to be backported ?

StefanRijnhart:
@rousseldenis I don't have time for that but it should be easy for the person who requires the module to work on 16.0

AaronHForgeFlow (approved):
Code LGTM! I guess we can keep the migration here until the 16.0 version is added

StefanRijnhart:
@AaronHForgeFlow That might never happen. Why wait?

AaronHForgeFlow:
I mean, merge this now and if for some reason it is added in 16.0 create a new PR to remove the migration scripts here.

StefanRijnhart:
@AaronHForgeFlow Thanks for clarifying! As for removing the script once a 16.0 module is published, we don't even have to do that as it would break the migration of people who migrate from 15 to 17 using the Odoo enterprise migration service like me. As it is, the migration script already checks if the field was created previously (in a future 16.0 version of this module) and then does nothing.

AaronHForgeFlow:
I know it is a pain for the ones that migrate with the enterprise service. But I think it is a "rule" to delete migrations folders of older versions when doing a migration (https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0). Perhaps you can open a separate discussion in regards that topic. I am not against merging this now, so I will not block the PR because of it.

StefanRijnhart:
@AaronHForgeFlow you are right, I will open that discussion if the matter becomes concrete again.

dreispt (approved):

PR #673, opened 2024-07-21
[16.0][ADD] analytic_mixin_analytic_accounts_and_plans

norlinhenrik:
This PR is based on #565. See also [PR 1207 for account_financial_report](https://github.com/OCA/account-financial-reporting/pull/1207).

`account.move.line` new fields:
- `analytic_account_ids` (useful for `account_financial_report`)
- `analytic_account_names` (introduced in #565)
- `analytic_plan_ids` (introduced by me)
- `analytic_plan_names` (for consistency, but I don't need it)

The compute method in `account_financial_report` (AGPL) is written for stored fields, with good performance.
The compute method in #565 (LGPL) is written for non-stored fields.

QUESTIONS
1) If `account_financial_report` depends on this module for `analytic_account_ids`, will the **performance** suffer?
2) If no dependency, and both modules introduce `analytic_account_ids`, can both be used in the same database?
3) What should be the module **name**?


pedrobaeza:
Please avoid plurals in module names, and I would say that this requires a shorter name that summarizes the goal of the module.

norlinhenrik:
I suggested "accounts" and "plans" since they are many2many fields. But I am fine with "account" and "plan".

What would be a good shorter name?

- `analytic_account_and_plan`?
- `analytic_account_and_plan_mixin`?
- `analytic_mixin_fields`?

pedrobaeza:
What about `analytic_unfolding_mixin`?

PR #674, opened 2024-07-25
[16.0][ADD] account_analytic_distribution_manual: New module to easily auto…

carlos-lopez-tecnativa:
…complete analytic accounts on any model that has a field for analytic accounts.

@Tecnativa TT50322

PR #675, opened 2024-07-26
[16.0][REM] account_analytic_tag_distribution: Remove view that is incompatible with other modules and hides native functionality

carlos-lopez-tecnativa:
The purpose of this module is to add a new option for using analytic accounts, but it should not hide the native analytic field.

Hiding this field causes tests to fail on PR  https://github.com/OCA/account-analytic/pull/674

@Tecnativa TT50322

@victoralmau @pedrobaeza could you please take a look?


pedrobaeza:
Included in #674

carlos-lopez-tecnativa:
Changes are made in the same PR referenced, so I'm closing this one. Sorry for opening another PR.

PR #677, opened 2024-08-07
[16.0][IMP] account_analytic_distribution_manual: Enhanced control of analytics account from manual distribution

carlos-lopez-tecnativa:
- Show the current distribution selected at the top of the tags
- When unselecting a distribution, all related analytic accounts are also unselected
- post_init_hook to convert old data to new models(if table exists)

@Tecnativa TT50235 
@carolinafernandez-tecnativa @pedrobaeza could you please take a look

image

### DIFF
--- a/account_analytic_distribution_manual/hooks.py
+++ b/account_analytic_distribution_manual/hooks.py
@@ -1,0 +1,56 @@
+# Copyright 2024 Tecnativa - Carlos Lopez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import SUPERUSER_ID, api, tools
+
+
+def post_init_hook(cr, registry):
+    if tools.table_exists(cr, "account_analytic_tag"):
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        sql = """
+        WITH counted_tags AS (
+            SELECT
+                tag.id,
+                tag.name,
+                tag.active,
+                tag.company_id,
+                ROW_NUMBER() OVER (PARTITION BY tag.name ORDER BY tag.id) AS row_count
+            FROM account_analytic_tag tag
+            WHERE tag.active_analytic_distribution = true
+        )
+        SELECT
+            CASE
+                WHEN row_count = 1 THEN tag.name
+                ELSE CONCAT(tag.name, ' (', tag.id, ')')
+            END AS name,
+            tag.id,
+            tag.active,
+            tag.company_id,
+            distribution.account_id,
+            distribution.percentage
+        FROM
+            counted_tags tag
+            INNER JOIN
+                account_analytic_distribution distribution ON tag.id = distribution.tag_id;
+
+        """
+        env.cr.execute(sql)
+        distribution_by_tag = {}
+        distribution_manual_vals = []
+        for data in env.cr.dictfetchall():
+            tag_key = (data["name"], data["active"], data["company_id"])
+            distribution_by_tag.setdefault(tag_key, []).append(data)
+        for tag_key, distributions in distribution_by_tag.items():
+            tag_name, tag_active, company_id = tag_key
+            distribution_manual_val = {
+                "name": tag_name,
+                "active": tag_active,
+                "company_id": company_id or env.company.id,
+                "analytic_distribution": {
+                    distribution["account_id"]: distribution["percentage"]
+                    for distribution in distributions
+                },
+            }
+            distribution_manual_vals.append(distribution_manual_val)
+        if distribution_manual_vals:
+            env["account.analytic.distribution.manual"].create(distribution_manual_vals)
--- a/account_analytic_distribution_manual/models/base.py
+++ b/account_analytic_distribution_manual/models/base.py
@@ -1,0 +1,86 @@
+# Copyright 2024 Tecnativa - Carlos Lopez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import json
+
+from lxml import etree
+
+from odoo import api, models
+from odoo.tools.misc import frozendict
+
+
+class AnalyticMixin(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model
+    def get_view(self, view_id=None, view_type="form", **options):
+        """
+        The purpose of inheriting this method is to
+        add a specific field, 'manual_distribution_id',
+        to the tree/form views if the model has this field defined.
+
+        Additionally, if a one2many field exists
+        and the related model has the 'manual_distribution_id' field,
+        the method will add the 'manual_distribution_id' field to the sub-view.
+
+        Finally, the method returns the modified view.
+
+        Note: This method should be inherited in the base model,
+        not in the analytic.mixin model,
+        to ensure it executes correctly for models
+        that do not inherit from the mixin but have one2many fields.
+        For example, when rendering account.move,
+        it does not inherit from analytic.mixin,
+        but it has one2many fields that inherit from analytic.mixin.
+        """
+
+        def add_field(node, view_type, res_model):
+            attribute = "column_invisible" if view_type == "tree" else "invisible"
+            modifiers = json.dumps({attribute: True})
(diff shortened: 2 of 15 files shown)


carolinafernandez-tecnativa:
Functional review LGTM :) Please check CI. Thanks

carolinafernandez-tecnativa:
If you add a line without product, and add manual distribution, and then proceed to add product, analytic accounts dissapear but the tag is still there, i think it should be removed

image

carolinafernandez-tecnativa:
If you configure an Analytic Distribution Models for that manual distribution:

image

When adding new invoice line, the tag is not being added:

image

carlos-lopez-tecnativa:
@carolinafernandez-tecnativa  I reviewed your comments. However, these changes are outside the scope of this module. I have updated the roadmap with these notes. If you have additional questions, please let me know.

carolinafernandez-tecnativa (approved):

account_analytic_distribution_manual/models/base.py
```diff
@@ -0,0 +1,66 @@
...
+
+from odoo import api, models
+from odoo.tools.misc import frozendict
+
+
+class AnalyticMixin(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model
+    def get_view(self, view_id=None, view_type="form", **options):
```
pedrobaeza, line 15: Why not doing this in the mixin? Overriding base model is too much and must be avoided.

account_analytic_distribution_manual/models/base.py
```diff
@@ -0,0 +1,66 @@
...
+        def add_field(node, view_type, res_model):
+            attribute = "column_invisible" if view_type == "tree" else "invisible"
+            modifiers = json.dumps({attribute: True})
```
pedrobaeza, line 38: Are you sure you need to add manually the modifiers? I haven't needed that on the past.

account_analytic_distribution_manual/models/base.py
```diff
@@ -0,0 +1,66 @@
...
+            }
+            field_element = etree.SubElement(node, "field", field_options)
+            new_arch, new_models = View.postprocess_and_fields(field_element, res_model)
+            _merge_view_fields(all_models, new_models)
+            return field_element
+
+        def model_has_field(model):
+            return manual_distribution_field_name in self.env[model]._fields
+
+        def _merge_view_fields(all_models, new_models):
```
pedrobaeza, line 51: This sounds weird and I didn't need all of that. Maybe you need to call a standard method though. Please check other examples of the new `get_view` override.

account_analytic_distribution_manual/models/base.py
celm1990: Because in the mixin, this code is only executed on models that inherit from this mixin. For example, the account.move model does not inherit from this mixin. Therefore, when processing views, it embeds a view into account.move.line, but this code is not executed. For this reason, I implemented the code in the base model. I check if the model has the field manual_distribution_id or if it has a one2many field where the related model has the field.

carlos-lopez-tecnativa: Because in the mixin, this code is only executed on models that inherit from this mixin. For example, the account.move model does not inherit from this mixin. Therefore, when processing views, it embeds a view into account.move.line, but this code is not executed. For this reason, I implemented the code in the base model. I check if the model has the field manual_distribution_id or if it has a one2many field where the related model has the field.

account_analytic_distribution_manual/models/base.py
carlos-lopez-tecnativa: Yes, it's necessary because the modifier for the tree view is different from other views. Otherwise, the field would be shown in the views.

account_analytic_distribution_manual/models/base.py
pedrobaeza: Please put all this explanation as comments in the code, and next time, anticipating that the decision is going to be controversial, please document it in advance.

account_analytic_distribution_manual/models/base.py
```diff
@@ -0,0 +1,66 @@
...
+                if model in all_models:
+                    all_models[model] = tuple(set(all_models[model]) | set(view_fields))
+                else:
+                    all_models[model] = tuple(view_fields)
+
+        result = super().get_view(view_id=view_id, view_type=view_type, **options)
+        if view_type in ["tree", "form"]:
+            View = self.env["ir.ui.view"]
+            manual_distribution_field_name = "manual_distribution_id"
+            all_models = result["models"].copy()  # {modelname(str) ➔ fields(tuple)}
```
pedrobaeza, line 63: Please restrict this processing the maximum and optimize it. We are injecting this code for every view...

account_analytic_distribution_manual/models/base.py
carlos-lopez-tecnativa: I took an example from [base_tier_validation](https://github.com/OCA/server-ux/blob/b79368067c5b9b19a6e2f4e40234433239074819/base_tier_validation/models/tier_validation.py#L714). Without this code, when I add a field to an embedded view, it raises an error, so it's necessary(mainly for one2many fields).

pedrobaeza: Well, I don't know if that code is the most optimal ones. Please check another overrides to compare.

account_analytic_distribution_manual/models/base.py
carlos-lopez-tecnativa: I don't understand what optimization I can implement. I need to check if the current model inherits from the mixin or if this model has a one2many field that inherits from the mixin. I don't see how this code can be more optimized. Can you please give an example?

pedrobaeza: For example, don't copy this, which is a memory drainer and not sure why do you need to do it.

carlos-lopez-tecnativa: This variable `result["models"]` is a `frozendict`, so I can't modify it without making a new copy.

account_analytic_distribution_manual/models/base.py
carlos-lopez-tecnativa: I have seen many similar cases where this function is inherited, but most modules only inherit the function to change `strings` or `modifiers`, not to `add new fields`. Here is a similar case that makes the same changes: [base_cancel_confirm](https://github.com/OCA/server-ux/blob/b79368067c5b9b19a6e2f4e40234433239074819/base_cancel_confirm/model/base_cancel_confirm.py#L62-L75).

If I remove this code, the error is as shown in the image below. Adding the missing attribute (in this case, a `string`) causes a new error because it lacks the field information handled by `postprocess_and_fields`, which returns new fields and models.
image

account_analytic_distribution_manual/models/base.py
carlos-lopez-tecnativa: done

account_analytic_distribution_manual/hooks.py
```diff
@@ -0,0 +1,36 @@
...
+                distribution.account_id, distribution.percentage
+            FROM account_analytic_distribution distribution
+                INNER JOIN account_analytic_tag tag ON tag.id = distribution.tag_id
+            WHERE active_analytic_distribution = true
+        """
+        env.cr.execute(sql)
+        distribution_by_tag = {}
+        distribution_manual_vals = []
+        for data in env.cr.dictfetchall():
+            tag_key = (data["name"], data["active"], data["company_id"])
```
pedrobaeza, line 41: In my tests, I have found that there are duplicated tag names between active and no active, provoking an unique name, but maybe the problem is wider, as it seems previous tags didn't have an unique name constraint, so this code is going to mix the contents of all the tags with the same name.

carlos-lopez-tecnativa: I updated the SQL. Please let me know if you have any comments.

With these tags:
image

The result obtained is:
image

pedrobaeza (approved):
/ocabot merge major

PR #679, opened 2024-08-14
[16.0][IMP] account_analytic_distribution_manual: change field type from integer to many2one

carlos-lopez-tecnativa:
- If the user needs to group by this field, it was impossible when it was an integer.
- Added a script to populate the new field in models related to account_analytic_tags. 
Note: When multiple tags are present, only the last record is populated.

@Tecnativa TT50235

@carolinafernandez-tecnativa  @pedrobaeza could you please review these changes?

### DIFF
--- a/account_analytic_distribution_manual/hooks.py
+++ b/account_analytic_distribution_manual/hooks.py
@@ -3,10 +3,112 @@
 
 from odoo import SUPERUSER_ID, api, tools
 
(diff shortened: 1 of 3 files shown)


carolinafernandez-tecnativa (approved):
Funcional review LGTM, Thanks :)

pedrobaeza (approved):
Tested on the real instance

/ocabot merge patch

PR #680, opened 2024-08-16
[16.0][FIX] account_analytic_required: remove unneeded dependency

sbidoul:
The dependency on `account_usability` is not required to install the module, ~and let it do it's validation purpose.~

~To configure it, Odoo EE users don't need anything else. In CE, accounting users install account_usability anyway.~

~So this commit makes it compatible with both editions.~

nor to configure it


sbidoul:
@sbejaoui @rousseldenis

rousseldenis (approved):

rvalyi (approved):

PR #681, opened 2024-08-16
[14.0][ADD] account_analytic_line_split: add new module

WesleyOliveira98:
Ticket HT00788

cc: @marcelsavegnago @kaynnan @douglascstd


douglascstd (changes requested):
[FUNCIONTAL REVIEW] [NEED TO CHECK]

1. Test to split some analytic item (OK)
image

2. It could be better with the amount field
image

3. The child_ids table is duplicated, maybe it's an effect if install other module: (account_analytic_wip). If the modules use same field it could be a problem in the future?
image
This field is created in [analytic_activity_based_cost](https://github.com/OCA/account-analytic/tree/924b249954cd05f3c14be994bb5a14c824aedf6a/analytic_activity_based_cost) module on:
https://github.com/OCA/account-analytic/blob/924b249954cd05f3c14be994bb5a14c824aedf6a/analytic_activity_based_cost/models/account_analytic_line.py#L19
I guess this module must to be included on dependency in manifest.py, is it make sense?

WesleyOliveira98:
> 2\. It could be better with the amount field

This tree view is added with `analytic activity based cost`

WesleyOliveira98:
> 3\. I guess this module must to be included on dependency in manifest.py, is it make sense?

This module is in Alpha, when I tried to depend on it this error was triggered
```
account_analytic_line_split (Beta) depends on analytic_activity_based_cost (Alpha)
```
And I think it doesn't make much sense to depend on this module because the purposes of the modules are different, I change the name of the fields to `parent_line_id` and `child_line_ids`, and in the view, I control the visibility of the fields only displaying them when they are filled in

WesleyOliveira98:
I also added unit tests, this pull request is ready for review now

WesleyOliveira98:
To review this pull request, you must uninstall `account_analytic_no_lines` because this module disables the generation of an analytic line from a move line

WesleyOliveira98:
ping @OCA/accounting-maintainers

marcelsavegnago:
ping @OCA/accounting-maintainers

PR #682, opened 2024-08-19
[17.0][REF] stock_analytic: Split test utilities into shareable base class

aisopuro:
Needed to also rename method starting with double-underscrore to avoid Python method name mangling.

Intended to fix #644

### DIFF
--- a/stock_analytic/tests/test_stock_picking.py
+++ b/stock_analytic/tests/test_stock_picking.py
@@ -12,7 +12,7 @@ from odoo.exceptions import ValidationError
 from odoo.tests.common import TransactionCase
 
 
-class TestStockPicking(TransactionCase):
+class CommonStockPicking(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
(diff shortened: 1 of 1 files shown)


stock_analytic/tests/test_stock_picking.py
```diff
@@ -12,7 +12,7 @@
 from odoo.tests.common import TransactionCase
 
 
-class TestStockPicking(TransactionCase):
+class CommonStockPicking(TransactionCase):
```
rousseldenis, line 15: Please put it in a separate `common.py` file

pedrobaeza: Please don't. That makes the blame a hell. It's not technically needed anyway, and I doubt its usefulness as organizational element.

pedrobaeza (approved):

aisopuro:
Please hold off on merging this for a few minutes, it seems this change did not actually fix the issue I was trying to fix.

StefanRijnhart:
Please change commit message to include the module name. `[REF] stock_analytic: split test utilities into shareable base class`

aisopuro:
> Please change commit message to include the module name. `[REF] stock_analytic: split test utilities into shareable base class`

Sorry for the wait @StefanRijnhart , this should be fixed now, and this PR is good to review and merge on my part.

StefanRijnhart (approved):
Thanks!

PR #684, opened 2024-08-27
[16.0] [IMP] account_analytic_distribution_manual: add field manual_distribution_id into account.invoice.report

carlos-lopez-tecnativa:
@Tecnativa TT50235

@pedrobaeza could you please take a look

### DIFF
--- a/account_analytic_distribution_manual/__init__.py
+++ b/account_analytic_distribution_manual/__init__.py
@@ -1,4 +1,5 @@
 # Copyright 2024 Tecnativa - Carlos Lopez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from . import models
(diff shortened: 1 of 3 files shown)


account_analytic_distribution_manual/__init__.py
```diff
@@ -1,4 +1,5 @@
 # Copyright 2024 Tecnativa - Carlos Lopez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from . import models
+from . import report
```
pedrobaeza: Please use plural here (the same as `models` or `wizards`):

```suggestion
from . import reports
```

pedrobaeza (approved):

victoralmau (approved):

PR #685, opened 2024-08-28
[16.0][IMP] account_analytic_distribution_manual: Add extra json field for export/import of analytical distributions

victoralmau:
Add extra json field for export/import of analytical distributions

@Tecnativa TT50235

### DIFF
--- a/account_analytic_distribution_manual/models/account_analytic_distribution_manual.py
+++ b/account_analytic_distribution_manual/models/account_analytic_distribution_manual.py
@@ -1,4 +1,5 @@
 # Copyright 2024 Tecnativa - Carlos Lopez
+# Copyright 2024 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import _, api, fields, models
 
@@ -10,6 +11,15 @@ class AccountAnalyticDistributionManual(models.Model):
 
     name = fields.Char(required=True)
     active = fields.Boolean(default=True)
+    analytic_distribution_import = fields.Json(
+        compute="_compute_analytic_distribution_import",
+        inverse="_inverse_analytic_distribution_import",
+        readonly=False,
+        string="Analytic distribution (importable)",
+        help="Defining this field, it will set the analytical distribution in JSON "
+        "format, but using the analytic accounts names as keys of the dictionary, so it "
+        "eases the human input.",
+    )
     company_id = fields.Many2one(
         "res.company", required=True, default=lambda self: self.env.company
     )
@@ -21,6 +31,34 @@ class AccountAnalyticDistributionManual(models.Model):
             "The name must be unique per Company!",
         ),
     ]
+
+    @api.depends("analytic_distribution")
+    def _compute_analytic_distribution_import(self):
+        aa_model = self.env["account.analytic.account"]
+        for item in self:
+            data = {}
+            distribution = item.analytic_distribution
+            for key in list(distribution.keys()):
+                aa_record = aa_model.browse(int(key))
+                data[aa_record.name] = distribution[key]
+            item.analytic_distribution_import = data
+
+    def _inverse_analytic_distribution_import(self):
+        """Convert the json to the appropriate value of analytic_distribution."""
+        aa_model = self.env["account.analytic.account"]
+        for item in self:
+            base_domain = [("company_id", "in", item.company_id.ids + [False])]
+            data = {}
+            new_distribution = item.analytic_distribution_import
+            for key in list(new_distribution.keys()):
+                domain = base_domain + [("name", "=", key)]
+                aa_record = aa_model.search(
+                    domain,
+                    limit=1,
+                )
+                if aa_record:
+                    data[aa_record.id] = new_distribution[key]
+            item.analytic_distribution = data
 
     @api.returns("self", lambda value: value.id)
     def copy(self, default=None):
--- a/account_analytic_distribution_manual/tests/test_analytic_distribution_manual.py
+++ b/account_analytic_distribution_manual/tests/test_analytic_distribution_manual.py
@@ -1,4 +1,5 @@
 # Copyright 2024 Tecnativa - Carlos Lopez
+# Copyright 2024 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 from psycopg2.errors import UniqueViolation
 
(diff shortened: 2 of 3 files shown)


victoralmau:
Ping @pedrobaeza What do you think of this approach?

account_analytic_distribution_manual/models/account_analytic_distribution_manual.py
```diff
@@ -10,6 +13,13 @@ class AccountAnalyticDistributionManual(models.Model):
 
     name = fields.Char(required=True)
     active = fields.Boolean(default=True)
+    analytic_distribution_text = fields.Text(
+        compute="_compute_analytic_distribution_text",
+        inverse="_inverse_analytic_distribution_text",
+        readonly=False,
+        string="Analytic distribution (text)",
+        help="Defining this text field will set the analytical distribution (json).",
```
pedrobaeza:
```suggestion
    analytic_distribution_text = fields.Text(
        compute="_compute_analytic_distribution_text",
        inverse="_inverse_analytic_distribution_text",
        readonly=False,
        string="Analytic distribution (importable)",
        help="Defining this text field will set the analytical distribution (json).",
```

pedrobaeza (commented):
I think it's simpler: use `json.dumps` and `json.loads` for compute and inverse.

victoralmau:
Your approach is not possible because the `analytic_distribution` field contains the **id** of the analytic accounts, something that is not useful to use in imports, so it is transformed to the `name` field.

account_analytic_distribution_manual/models/account_analytic_distribution_manual.py
```diff
@@ -22,6 +32,29 @@ class AccountAnalyticDistributionManual(models.Model):
...
+                aa_record = aa_model.browse(int(key))
+                data[aa_record.name] = distribution[key]
+            item.analytic_distribution_text = str(data)
+
+    def _inverse_analytic_distribution_text(self):
+        """Convert the json text to the appropriate value of analytic_distribution."""
+        aa_model = self.env["account.analytic.account"]
+        for item in self:
+            data = {}
+            new_distribution = ast.literal_eval(item.analytic_distribution_text)
```
pedrobaeza: This may involve security risks. Can you use `json.loads` here?

account_analytic_distribution_manual/tests/test_analytic_distribution_manual.py
```diff
@@ -24,3 +25,39 @@ def test_copy_manual_distribution(self):
                     "name": "Manual Distribution 1",
                 }
             )
+
+    def test_manual_distribution_analytic_distribution_text(self):
+        aa_1 = self.analytic_account_a1
+        aa_2 = self.analytic_account_a2
+        self.assertEqual(
+            self.distribution_1.analytic_distribution_text,
+            str({aa_1.name: 40.0, aa_2.name: 60.0}),
```
pedrobaeza: You must hardcode this result for proper test asserts. If not, you are creating business code inside the test code.

pedrobaeza: I mean:

```suggestion
            '{"blabla": 40.0, "blabla2": 60.0}',
```

victoralmau:
I have changed it to a json field because it works correctly (reducing code) and the data is exported and imported correctly.

pedrobaeza:
And do you know why the standard one is not importable?

https://github.com/odoo/odoo/blob/1d74864ecb0fd922d23d0da2d32f99c67d1dddbf/addons/analytic/models/analytic_mixin.py#L11

Maybe because the absence of the `inverse`? It's a computed writable.

victoralmau:
> And do you know why the standard one is not importable?
> 
> https://github.com/odoo/odoo/blob/1d74864ecb0fd922d23d0da2d32f99c67d1dddbf/addons/analytic/models/analytic_mixin.py#L11
> 
> Maybe because the absence of the `inverse`? It's a computed writable.

The main problem with using the `analytic_distribution` field is what it contains 
`{1: 40.0, 2: 60:0}`
this is not “_readable_” for importing data, for this reason the `analytic_distribution_import` field has been added.
`{“Analytic account A”: 40.0, “Analytic account B”: 60:0}`

account_analytic_distribution_manual/models/account_analytic_distribution_manual.py
```diff
@@ -10,6 +11,13 @@ class AccountAnalyticDistributionManual(models.Model):
 
     name = fields.Char(required=True)
     active = fields.Boolean(default=True)
+    analytic_distribution_import = fields.Json(
+        compute="_compute_analytic_distribution_import",
+        inverse="_inverse_analytic_distribution_import",
+        readonly=False,
+        string="Analytic distribution (importable)",
+        help="Defining this field will set the analytical distribution (json).",
```
pedrobaeza: Include here an explanation about using this field for searching the analytic accounts by name.

pedrobaeza (commented):
OK, then let's continue with this, only adding the explanation in the help.

account_analytic_distribution_manual/models/account_analytic_distribution_manual.py
```diff
@@ -10,6 +11,14 @@ class AccountAnalyticDistributionManual(models.Model):
 
     name = fields.Char(required=True)
     active = fields.Boolean(default=True)
+    analytic_distribution_import = fields.Json(
+        compute="_compute_analytic_distribution_import",
+        inverse="_inverse_analytic_distribution_import",
+        readonly=False,
+        string="Analytic distribution (importable)",
+        help="""Defining this field will set the analytical distribution (json). It is
+        used this field for searching the analytic accounts by name.""",
```
pedrobaeza:
```suggestion
        help="""Defining this field, it will set the analytical distribution in JSON format, but using the analytic accounts names as keys of the dictionary, so it eases the human input.""",
```

account_analytic_distribution_manual/models/account_analytic_distribution_manual.py
```diff
@@ -10,6 +11,15 @@ class AccountAnalyticDistributionManual(models.Model):
...
     name = fields.Char(required=True)
     active = fields.Boolean(default=True)
+    analytic_distribution_import = fields.Json(
+        compute="_compute_analytic_distribution_import",
+        inverse="_inverse_analytic_distribution_import",
+        readonly=False,
+        string="Analytic distribution (importable)",
+        help="""Defining this field, it will set the analytical distribution in JSON
+        format, but using the analytic accounts names as keys of the dictionary, so it
+        eases the human input.""",
```
pedrobaeza: I don't think you should use the triple quote, as this introduces the line feeds and spaces in the text and translations:

```suggestion
        help="Defining this field, it will set the analytical distribution in JSON "
        "format, but using the analytic accounts names as keys of the dictionary, so it "
        "eases the human input.",
```

pedrobaeza (approved):

victoralmau:
Please @carlos-lopez-tecnativa can you review it?

account_analytic_distribution_manual/models/account_analytic_distribution_manual.py
```diff
@@ -22,6 +32,29 @@ class AccountAnalyticDistributionManual(models.Model):
...
+            item.analytic_distribution_import = data
+
+    def _inverse_analytic_distribution_import(self):
+        """Convert the json to the appropriate value of analytic_distribution."""
+        aa_model = self.env["account.analytic.account"]
+        for item in self:
+            data = {}
+            new_distribution = item.analytic_distribution_import
+            for key in list(new_distribution.keys()):
+                aa_record = aa_model.search([("name", "=", key)], limit=1)
```
carlos-lopez-tecnativa:
```suggestion
                aa_record = aa_model.search([("name", "=", key), ("company_id", "=", self.env.company.id)], limit=1)
```

I think it is better to add `company_id` to the domain to avoid selecting accounts from other companies (or at least minimize the problem). I created two records with the same name in different companies. The manual distribution belongs to San Francisco. I selected both companies, imported the file, and it picked the account from the other company. Then, I selected only one company and got an error.

image

image


image

carlos-lopez-tecnativa (changes requested):
Functional testing, only a few comments to avoid errors in multi-company. What do you think?

account_analytic_distribution_manual/models/account_analytic_distribution_manual.py
victoralmau: Thanks, changes done

account_analytic_distribution_manual/models/account_analytic_distribution_manual.py
```diff
@@ -22,6 +32,34 @@ class AccountAnalyticDistributionManual(models.Model):
...
+
+    def _inverse_analytic_distribution_import(self):
+        """Convert the json to the appropriate value of analytic_distribution."""
+        aa_model = self.env["account.analytic.account"]
+        base_domain = [("company_id", "in", self.env.company.ids + [False])]
+        for item in self:
+            data = {}
+            new_distribution = item.analytic_distribution_import
+            for key in list(new_distribution.keys()):
+                domain = base_domain + [("name", "=", key)]
```
carlos-lopez-tecnativa: Sorry, I see we can access the company on each `item`, so it’s better to use this field instead of the environment company.

```suggestion
        for item in self:
            base_domain = [("company_id", "in", item.company_id.ids + [False])]
            data = {}
            new_distribution = item.analytic_distribution_import
            for key in list(new_distribution.keys()):
                domain = base_domain + [("name", "=", key)]
```

carlos-lopez-tecnativa (approved):
LGTM, thanks

PR #687, opened 2024-09-05
[FEAT] mrp_stock_analytic: add analytic to new raw lines 

aisopuro:
Depends on:
* #644 

Built on top of the dependent PR, only the last commit is actually different.


aisopuro:
Rebased and re-pushed. I would like to request this be un-staled.

PR #688, opened 2024-09-09
[16.0][IMP] account_analytic_distribution_manual: Add manual_distribution_id field to analytic items

victoralmau:
Add `manual_distribution_id` field to analytic items

Please @pedrobaeza and  @carlos-lopez-tecnativa can you review it?

@Tecnativa TT50751

### DIFF
--- a/account_analytic_distribution_manual/hooks.py
+++ b/account_analytic_distribution_manual/hooks.py
@@ -182,3 +182,14 @@ def post_init_hook(cr, registry):
                         """,
                         (distribution_map[tag_id].id, res_id),
                     )
+        # Define the value of manual_distribution_id in the line items
+        env.cr.execute(
+            """
+            UPDATE account_analytic_line AS aal
+            SET manual_distribution_id = aml.manual_distribution_id
+            FROM account_move_line AS aml
+            WHERE aal.move_line_id = aml.id
+            AND aml.manual_distribution_id IS NOT NULL
+            AND aal.manual_distribution_id IS NULL
+            """,
+        )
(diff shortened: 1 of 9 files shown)


account_analytic_distribution_manual/migrations/16.0.2.3.0/post-migration.py
```diff
@@ -0,0 +1,19 @@
...
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_analytic_line AS aal
```
pedrobaeza, line 12: This should be also on post_init_hook

victoralmau: Do you mean that in post_init_hook the `account.analytic.line` that have `account_analytic_tag` should be linked with the corresponding `manual_distribution_id`? That will only work if there is already account analytic distribution manual created and they have exactly the **same** name as the analytic tag (because they are not linked in any way).

pedrobaeza: The post init hook creates the manual distributions, so I don't get you. We have to link the existing data.

account_analytic_distribution_manual/hooks.py
```diff
@@ -157,6 +157,18 @@ def post_init_hook(cr, registry):
             new_distribution = DistributionManual.create(distribution_manual_val)
             distribution_map[tag_id] = new_distribution
             all_tag_ids.append(tag_id)
+            # Define the value of manual_distribution_id in the line items
+            update_sql = f"""
+            UPDATE account_analytic_line
+            SET manual_distribution_id = {new_distribution.id}
+            WHERE manual_distribution_id IS NOT NULL
+            ANBD id IN (
```
pedrobaeza: This contains a typo and is not correct, as `manual_distribution_id` will always be NULL in this state.

pedrobaeza: Why not using the same expression as the migration script?

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #691, opened 2024-09-18
[16.0][ADD] account_analytic_organization: New module account_analytic_organization

mpascuall:
This module adds a new field on the partner. This organization field is also added on the account analytic line and account move line so you can filter and group by it.

cc https://github.com/APSL 160524
@miquelalzanillas @lbarry-apsl @javierobcn @peluko00 @BernatObrador @ppyczko please review


peluko00 (approved):
Good work, tested in runboat and code review. LGTM!

miquelalzanillas:
A filter for 'analytic_org_id' field in both search views of journal and analytic entries would be great. Please, can you add it?

image

mpascuall:
> A filter for 'analytic_org_id' field in both search views of journal and analytic entries would be great. Please, can you add it?
> 
> image

Thanks for your feedback! I added the filter in both views

miquelalzanillas:
Hello again @mpascuall ,

I missed to check the form view for the new model 'account.analytic.organization' in the previous review. 

As this new model has a single field ('name') I think would be much useful using inline edition in the list view instead of using the form view.  What do you think?

mpascuall:
> Hello again @mpascuall ,
> 
> I missed to check the form view for the new model 'account.analytic.organization' in the previous review.
> 
> As this new model has a single field ('name') I think would be much useful using inline edition in the list view instead of using the form view. What do you think?

Hi! You're absolutely right, thanks for the feedback! Could you please review again?

miquelalzanillas (approved):
Code and functional review. LGTM.

miquelalzanillas:
Hi @pedrobaeza ,

Could you review please?

ppyczko (approved):
LGTM, tested in runboat

PR #693, opened 2024-09-23
[16.0][IMP] account_analytic_distribution_manual: Add extra json field for export/import to analytic.mixin

victoralmau:
Related to https://github.com/OCA/account-analytic/pull/685

Add extra json field for export/import to `analytic.mixin`

Move the functionality to the `analytic.mixin` model so that it can also be used in `account.move.line` records

@Tecnativa TT50944

### DIFF
--- a/account_analytic_distribution_manual/models/analytic_mixin.py
+++ b/account_analytic_distribution_manual/models/analytic_mixin.py
@@ -1,9 +1,50 @@
 # Copyright 2024 Tecnativa - Carlos Lopez
+# Copyright 2024 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class AnalyticMixin(models.AbstractModel):
     _inherit = "analytic.mixin"
 
     manual_distribution_id = fields.Many2one("account.analytic.distribution.manual")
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

victoralmau:
Ping @carlos-lopez-tecnativa

account_analytic_distribution_manual/models/analytic_mixin.py
```diff
@@ -1,9 +1,47 @@
...
+            for key in list(distribution.keys()):
+                aa_record = aa_model.browse(int(key))
+                data[aa_record.name] = distribution[key]
+            item.analytic_distribution_import = data
+
+    def _inverse_analytic_distribution_import(self):
+        """Convert the json to the appropriate value of analytic_distribution."""
+        aa_model = self.env["account.analytic.account"]
+        for item in self:
+            base_domain = [("company_id", "in", item.company_id.ids + [False])]
```
carlos-lopez-tecnativa: I think this domain now depends on whether the model has a `company_id` field. It's not 100% reliable that the model will have this field.

```suggestion
            company = self.env.company
            if "company_id" in item._fields:
                company = item.company_id
            base_domain = [("company_id", "in", company.ids + [False])]
```

carlos-lopez-tecnativa (approved):
LGTM

PR #694, opened 2024-09-30
[16.0][ADD] account_analytic_document_date: New module account_analyt…

BernatObrador:
This module adds an other date to the account.move for analytics proposits

Usage:

1. **Set Document Date on Invoice:**
   - Navigate to the invoice and set the document date. 
   - If the document date is not set, it will automatically be assigned the invoice date upon confirmation.

2. **Analytic Line Generation:**
   - If any line has analytic information, the generated analytic entry will inherit the document date.

3. **Setting Document Date in Reconcile View:**
   - You can also set the document date from the reconciliation view, but this option is available only during manual reconciliation.

cc https://github.com/APSL 160768
@miquelalzanillas @lbarry-apsl @javierobcn @peluko00 @ppyczko please review


peluko00 (approved):
LGTM, code review and tested in runboat

ppyczko (approved):
LGTM, tested in runboat

BernatObrador:
@pedrobaeza Could you review please?

account_analytic_document_date/i18n/ca_ES.po
pedrobaeza, line 1: It should be `ca.po`

pedrobaeza (commented):
I don't understand the goal of having another date. I would understand that you want a different date for the analytic line, but not having dual dates.

BernatObrador:
@pedrobaeza 
The module introduces a new field in the accounting records to specify the "Accrual Date" for income and expenses. This field is crucial for accurately tracking when the revenue or expense is earned or incurred, regardless of the invoice creation date.

For instance, if an invoice is created in December but refers to sales made in November, the Accrual Date allows us to reflect this correctly in our accounting records. This ensures compliance with accrual-based accounting principles and provides better insights into the financial period to which transactions belong.

This feature is particularly valuable for businesses that need to differentiate between the accounting period of transaction occurrence and the timing of document issuance.

account_analytic_document_date/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    "name": "Account Analytic Document Date",
+    "summary": "Account Analytic Document Date",
+    "version": "16.0.1.0.0",
+    "website": "https://github.com/OCA/account-analytic",
+    "author": (
+        "(Nagarro - APSL), Miquel Pascual, Bernat Obrador,"
+        "Odoo Community Association (OCA)"
+    ),
+    "maintainers": ["mpascual@apsl.net, bobrador@apsl.net"],
```
pedrobaeza: These are Github logins, not emaisl.

account_analytic_document_date/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2024 (APSL - Nagarro) Miquel Pascual, Bernat Obrador
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Account Analytic Document Date",
+    "summary": "Account Analytic Document Date",
+    "version": "16.0.1.0.0",
+    "website": "https://github.com/OCA/account-analytic",
+    "author": (
+        "(Nagarro - APSL), Miquel Pascual, Bernat Obrador,"
```
pedrobaeza: This is the company, not the persons

account_analytic_document_date/static/description/icon.svg
pedrobaeza, line 1: Don't put this if it's not modified from the standard one (which is the case).

account_analytic_document_date/models/account_move.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2024 (APSL - Nagarro) Miquel Pascual, Bernat Obrador
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    document_date = fields.Date()
```
pedrobaeza: If it's a date only for analytic, it shouldn't be called `document_date`, but `analytic_document_date` or `analytic_date`.

pedrobaeza (changes requested):

account_analytic_document_date/models/account_move.py
BernatObrador: Indeed, it is used for analytic purposes, but it is also visible in the account module, so it serves both purposes. For this reason, I believe the name 'document_date' remains appropriate.

pedrobaeza: But it creates confusion on users thinking that date is going to be used for example for financial reports. It should be renamed to `analytic_document_date` IMO.

BernatObrador:
@pedrobaeza Could you please review again?

account_analytic_document_date/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2024 (APSL - Nagarro) Miquel Pascual, Bernat Obrador
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Account Analytic Document Date",
+    "summary": "Account Analytic Document Date",
+    "version": "16.0.1.0.0",
+    "website": "https://github.com/OCA/account-analytic",
+    "author": ("(Nagarro - APSL), Odoo Community Association (OCA)"),
```
pedrobaeza: Is this the usual form (between parenthesis) of your company name?

pedrobaeza (approved):
/ocabot merge nobump

PR #701, opened 2024-10-10
[16.0][ADD] contract_analytic_tag: New module

victoralmau:
New module

@Tecnativa TT51221


carlos-lopez-tecnativa:
I think it's necessary to adapt to the new analytic distribution model instead of replicating the older one using analytic tags. Am I missing any context?

pedrobaeza:
In this case, we need this module because the analytic tag is used for classifying analytic lines, not for doing distributions, so the new model doesn't work for this (nor the manual distribution).

carlos-lopez-tecnativa (approved):
@pedrobaeza thanks for clarifying. 
Tested, LGTM

contract_analytic_tag/models/contract_line.py
```diff
@@ -0,0 +1,19 @@
...
+    analytic_tag_ids = fields.Many2many(
+        comodel_name="account.analytic.tag",
+        string="Analytic Tags",
+        domain="['|', ('company_id', '=', False), ('company_id', '=', company_id)]",
+    )
+
+    def _prepare_invoice_line(self):
+        vals = super()._prepare_invoice_line()
+        if self.analytic_tag_ids:
+            vals.update({"analytic_tag_ids": [(6, 0, self.analytic_tag_ids.ids)]})
```
pedrobaeza, line 18: Simpler syntax:

```suggestion
            vals["analytic_tag_ids"] = [(6, 0, self.analytic_tag_ids.ids)]
```

pedrobaeza (changes requested):
This should go to OCA/contract

victoralmau:
Moved to https://github.com/OCA/contract/pull/1133

PR #705, opened 2024-10-22
[16.0][IMP] account_move_update_analytic: suppress invoice sync

hbrunn:
depending on the set of modules installed, updating the analytic account might trigger a call to `_sync_dynamic_lines` which we don't want here


Shide (approved):

hbrunn:
superseded by https://github.com/OCA/account-analytic/pull/726

PR #726, opened 2024-12-10
[16.0][FIX] account_move_update_analytic: Update analytic tax distribution

hbrunn:
If you have a tax with the analytic flag enabled and change a line's analytic distribution with that tax, you'll depending on the odoo receive an error message or an empty analytic distribution without this patch.

I took the opportunity to add extensive tests for the module.

### DIFF
--- a/account_move_update_analytic/__init__.py
+++ b/account_move_update_analytic/__init__.py
@@ -1,1 +1,2 @@
+from . import models
 from . import wizards
--- a/account_move_update_analytic/models/account_move_line.py
+++ b/account_move_update_analytic/models/account_move_line.py
@@ -9,6 +9,14 @@ force_state_sentinel = object()
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
+    def _check_reconciliation(self):
+        """
+        Pass the check if we're asked to
+        """
+        if self.env.context.get("account_move_update_analytic") == force_state_sentinel:
+            return
+        return super()._check_reconciliation()
+
     def _compute_all_tax(self):
         """
         super() doesn't write the analytic distribution when the move is posted.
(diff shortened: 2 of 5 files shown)


remi-filament (changes requested):
Thanks @hbrunn for the extra tests and the fix.

I have 2 issues when testing on runboat : 

1. (not related to this module) A JS error which comes from account_analytic_distribution_manual module when all modules from this repo are installed : when I select a new distribution as before in the update wizard, I get a 
TypeError: this.fields[fieldName] is undefined
    update@http://oca-account-analytic-16-0-74a4fdd973b5.runboat.odoo-community.org/web/assets/802-c885ccf/web.assets_backend.min.js:2381:368

1. When testing with an invoice with 2 lines on the same tax (with analytic boolean set to True), if you validate your invoice without analytic distribution on any line you end up with only one line of tax (without analytic distribution). Now if I want to assign to each product line a different analytic distribution using this module wizard, only the first one will be reflected on the tax. When you do that before validating the invoice, Odoo would split the tax line into 2 tax lines : 1 per product with corresponding analytic distribution.

The proper solution here would be to split tax account.move.line per product as per Odoo default behaviour but then we are not only updating analytic but also account move, which I am not sure is a good idea !

Therefore I am not sure what is better : to not have the fix and update tax line manually, or to have the fix and not acknowledging that some distribution are not properly applied on taxes ?

hbrunn:
I think we need to do the tax line splitting after all - if you manually assign some analytic distribution to the tax line, it will be wrong too. How about for now allowing only the simple case (=there's no other line with the same tax in the move), raising for the complex case and adding a TODO in the readme?

remi-filament:
> I think we need to do the tax line splitting after all - if you manually assign some analytic distribution to the tax line, it will be wrong too. How about for now allowing only the simple case (=there's no other line with the same tax in the move), raising for the complex case and adding a TODO in the readme?

You could still recalculate it on your own and assign manually correctly, but yes, it makes sense as a first step to manage simple case where there tax belongs only to the line being modified, showing a Warning for the other cases !

hbrunn:
~~@remi-filament my customer needs this case after all, so I'll set this to draft until I come up with a solution~~

after quite some tinkering I got it to work

PR #729, opened 2024-12-16
[ADD] account_analytic_product: New module account_analytic_product

mpascuall:
This module adds product_id to journal entries

New module account_analytic_product

cc https://github.com/APSL 164801

@miquelalzanillas @lbarry-apsl @BernatObrador @peluko00 @javierobcn @ppyczko please review


pedrobaeza:
I think this has nothing to do with analytic, but general accounting.

mpascuall:
> I think this has nothing to do with analytic, but general accounting.

Yes, closing this pr and moving it into account-invoicing

pedrobaeza:
Better into account-financial-tools, and the name should reflect the goal, so the word `analytic` should go out.

PR #734, opened 2025-01-16
[16.0][FIX] pos_analytic_by_config: Avoid prefixing

chienandalu:
Prefixing the distribution shouldn't be necessary. We can hook into the sales move lines prepare method to add or session info that later can be used to figure out the proper distribution.

cc @Tecnativa TT52559

please review @pedrobaeza @victoralmau 

Some doubts that come to my mind:

- The new technique is a little bit hackish...
- With the former approach, we could match any account from the closing session using the prefixes. Would it make any sense anyway?

### DIFF
--- a/pos_analytic_by_config/migrations/16.0.1.0.0/post-migration.py
+++ b/pos_analytic_by_config/migrations/16.0.1.0.0/post-migration.py
@@ -26,18 +26,6 @@ def migrate(env, version):
     pos_configs = env["pos.config"].browse([pc.id for pc in pos_configs_dict.keys()])
     for company in pos_configs.company_id:
         company_configs = pos_configs.filtered(lambda x: x.company_id == company)
-        default_account_revenue = env["account.account"].search(
-            [
-                ("company_id", "=", company.id),
-                ("account_type", "=", "income"),
-                (
-                    "id",
-                    "!=",
-                    company.account_journal_early_pay_discount_gain_account_id.id,
-                ),
-            ],
-            limit=1,
-        )
         analytic_plan = env["account.analytic.plan"].create(
             {
                 "name": "Stores",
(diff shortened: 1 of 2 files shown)


pos_analytic_by_config/models/account_move_line.py
```diff
@@ -0,0 +1,18 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        """We need to receive this key from the vals as the create call is encampsulated
```
pedrobaeza:
```suggestion
        """We need to receive this key from the vals as the create call is encapsulated
```

pos_analytic_by_config/models/account_move_line.py
```diff
@@ -0,0 +1,18 @@
...
+        inside the _get_sale_vals method of the pos.session model."""
```
pedrobaeza: This is not explaining how and where `pos_config_id` context is used. docstring also puts the trailing `"""` in a new line according PEP257.

pedrobaeza (approved):
Let's avoid to lose this.

/ocabot merge patch

pedrobaeza:
Finally only the change in the migration script is needed, and the modified test is assuring that.

/ocabot merge patch

pedrobaeza:
It seems one tour is failing randomly:

/ocabot merge patch

pedrobaeza:
/ocabot merge patch

@carlos-lopez-tecnativa I'm having troubles with the tours of `account_analytic_distribution_manual`.

carlos-lopez-tecnativa:
> /ocabot merge patch
> 
> @carlos-lopez-tecnativa I'm having troubles with the tours of `account_analytic_distribution_manual`.

I ran the test locally but I couldn't reproduce the issue.
image

When I ran it a second time, I got two errors, but they were in other modules.

image

image

image

Please retry again

pedrobaeza:
@carlos-lopez-tecnativa now I'm reproducing the CI fail in the PR after forcing a rebuild. Please check.

PR #737, opened 2025-01-21
[MIG] purchase_analytic: Migration to 18.0

CLaurelB:
The invisible field in the purchase order form view was removed, as it will be added automatically if required [1].

[1]: https://github.com/odoo/odoo/pull/137031


CLaurelB:
Hello @luisg123v @StefanRijnhart

Could you please review this PR?

Best regards.

purchase_analytic/views/purchase_views.xml
```diff
@@ -0,0 +1,58 @@
...
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="arch" type="xml">
+            <field name="partner_ref" position="after">
+                <field
+                    name="analytic_distribution"
+                    widget="analytic_distribution"
+                    groups="analytic.group_analytic_accounting"
+                    readonly="state in ('done', 'cancel')"
+                />
+                <field name="analytic_distribution" invisible="1" />
```
StefanRijnhart: The rule in Odoo 18 is: you can remove fields that are always invisible, because if necessary, Odoo will add it automatically in the view. But I am not sure if this works with base_view_inheritance_extension. Can you check, and if the field can be removed, remove it. If not, please mention this in an XML comment.

StefanRijnhart (commented):
Thanks for taking this another version further so quickly!

purchase_analytic/views/purchase_views.xml
CLaurelB: You're right, the field could be removed. I have just removed it and documented it in the commit description.

StefanRijnhart: Perfect, thanks!

StefanRijnhart (approved):

luisg123v (approved):
LGTM :+1: 

Note:

I noticed there's no commit for precommit. If I run precommit before the migration commit, no changes are performed, so I assume it's expected.

rvalyi (approved):

PR #746, opened 2025-02-20
[17.0][ADD] stock_quant_analytic: analytic on inventory adjustment

RLeeOSI:
Extends stock_analytic to allow analytic distribution on inventory adjustments (stock.quant)


max3903 (approved):

luisDIXMIT (commented):
I tried in local but analytic_distribution is't propagate between stock.quant and stock.move. Am I doing something wrong?

PR #752, opened 2025-03-10
[15.0][FIX] account_analytic_tag_default: analytic_tag_ids is computed in standard

LoisRForgeFlow:
This was overlook during migration. As the module may be already installed in some instances, we corrected at least the inheritance problem, but it is possible to face weird interaction if used in conjunction with account.analytic.defaults. It is recommended to move tag defaults defined with this module to analytic default rules (account.analytic.defaut) and uninstall.


JoanSForgeFlow (approved):
Functional and Code Review. LGTM!

AaronHForgeFlow (approved):
It makes sense

PR #757, opened 2025-04-01
[16.0][ADD] analytic_hr_department_restriction: New module

victoralmau:
New module

@Tecnativa TT55725


victoralmau:
It is now ready to review and merge.

analytic_hr_department_restriction/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2025 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Analytic Hr department Restriction",
```
pedrobaeza: Please ellaborate more the description of the modules, not to put the technical name in capitals (and in fact, `Hr` is incorrect, in any case it should be `HR`):

```suggestion
    "name": "Analytic distributions restriction per HR department",
```

analytic_hr_department_restriction/models/account_analytic_account.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class AccountAnalyticAccount(models.Model):
+    _inherit = "account.analytic.account"
+
+    department_id = fields.Many2one(
+        comodel_name="hr.department",
+        string="Department",
+        domain="['|',('company_id', '=?', company_id),('company_id', '=', False)]",
```
pedrobaeza, line 13: Isn't this obtained with the `check_company=True`? Or at least, it should have both.

analytic_hr_department_restriction/security/security_rules.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="group_analytic_accounting_without_department" model="res.groups">
```
pedrobaeza: AFAIK, this is working because Odoo itself is not adding any other record rules, but if this changes in other subsequent modules or in the core, this won't work, being record rules with groups additive. Can you please do this overriding the ORM code like `sales_team_security` and other similar modules?

pedrobaeza (changes requested):

analytic_hr_department_restriction/models/account_analytic_account.py
victoralmau: It is not possible, `check_company=True` only works correctly when the `company_id` field in the corresponding model is required, and in the departments (`hr.department`), the `company_id` field is not required.

analytic_hr_department_restriction/security/security_rules.xml
victoralmau: Sorry, but I don't understand what the problem is in this case, i.e. we want to limit the analytical accounts/plans that are displayed. If in the future some other module (or core) adds some restriction (for example with followers) I don't think you should change that behavior, I think you should add this condition as an additive rule and not completely redefine the rules by “omitting” the rules that already exist, but it's just my opinion.

analytic_hr_department_restriction/readme/USAGE.md
```diff
@@ -0,0 +1,5 @@
+#. Go to an analytical plan/account and define your employee's department.
+#. Go to an analytical plan/account and define a different department than your employee's department.
+#. Go to an analytical plan/account and leave the department empty.
+#. Users without the "Invoicing / Billing" permission will be able to see only the plans/analytical accounts of the user's employee's department.
+#. Users with the "Analytic Accounting (without department)" will be able to view the plans/analytical accounts without a department, as will users with the permission "Invoicing / Billing".
```
carlos-lopez-tecnativa: Please render this file correctly.
image

analytic_hr_department_restriction/views/account_analytic_account_views.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_account_analytic_account_form" model="ir.ui.view">
+        <field name="model">account.analytic.account</field>
+        <field name="inherit_id" ref="analytic.view_account_analytic_account_form" />
+        <field name="arch" type="xml">
+            <field name="plan_id" position="before">
+                <field name="department_id" groups="account.group_account_invoice" />
```
carlos-lopez-tecnativa, line 8: I noticed a duplicate field.
I'm not sure if this module can be added as a dependency. What do you think?
https://github.com/OCA/account-analytic/blob/5a0f5ca9a23a3b875f03469ced7b0d9730eb2e2a/analytic_base_department/models/analytic.py#L12

image

carlos-lopez-tecnativa (changes requested):

analytic_hr_department_restriction/views/account_analytic_account_views.xml
victoralmau: I think it should not have the `analytic_base_department` dependency because for example, it adds fields that we don't need to the analytic items.

pedrobaeza: I agree we shouldn't add that dependency.

analytic_hr_department_restriction/security/security_rules.xml
pedrobaeza: But this is not how record rules with groups work. When there are several record rules, the result is the sum of both recordset results, which is not what we want. We always want to show only the department|empty ones, no matter if other modules change this.

victoralmau: I still do not understand the problem, i.e. the behavior is as follows:
- Users without “extra” permissions: will only see analytic plans/accounts that have department defined and is the department of the user's employee.
- Users with the “Analytic Accounting (without department)” permission: Only if you have this permission, you will additionally be able to view plans/accounts that have NO department defined.
- Users with the Invoking/Billing permission: Will be able to view all analytic plans/accounts without restriction.

pedrobaeza (approved):

carlos-lopez-tecnativa (commented):
Tested and working fine. I just found an edge case, please let me know if this case can be handled.
When an employee changes department, the record rules don't take the new department into account, and the filter is applied using the old department. I think it's necessary to invalidate the cache in this case. What do you think?

victoralmau:
> Tested and working fine. I just found an edge case, please let me know if this case can be handled. When an employee changes department, the record rules don't take the new department into account, and the filter is applied using the old department. I think it's necessary to invalidate the cache in this case. What do you think?

I have tried to replicate the use case you indicate and it works correctly.

Example:
- I created a department 1 and department 2.
- I created an analytical account 1 linked to department 1.
- I created an analytical account 2 linked to department 2.
- I modified the user Marc Demo to have analytical accounting permissions but not full accounting permissions.
- I logged in with user Marc Demo and did not see analytical account 1 nor analytical account 2 (correct because marc demo's department was different).
- I modified from the user Mitchell Admin the department of the Marc Demo employee to Department 1.
- Marc Demo did see the analytic account 1 but NOT the analytic account 2 (this is the expected behavior).
- I modified from user Mitchell Admin the department of Marc Demo's employee to Department 2.
- Marc Demo did see analytical account 2 but NOT analytical account 1 (this is the expected behavior).

carlos-lopez-tecnativa:
> I have tried to replicate the use case you indicate and it works correctly.
> 

I tested with this configuration:

- Analytic plan without department
- A1 with D1
- A2 with D2

At first, the Demo user has `D1` and sees `A1` , that's OK.
Then I changed the department from `D1` to `D2`.
After refreshing the window, I expected to see `A2`, but `A1` is still displayed.

The Demo user only has the following groups:

- Purchase User
- Analytic Accounting
- Analytic Accounting (without department)

I uploaded a video to show my case, please let me know if I'm missing any configuration.

https://github.com/user-attachments/assets/2321571e-92c5-417a-b5c3-46d16e3a84dc

victoralmau:
I have now tested locally the behavior you indicate but I have not been able to reproduce it, i.e. it works correctly.

ejemplo

carlos-lopez-tecnativa (approved):
LGTM, tested locally

PR #763, opened 2025-04-14
17.0_MIG_hr_expense_analytic_tag

iTecan:
Migrated from version 16.0.
Added label view in Expense Reports (was not on expense lines)


iTecan:
I don't understand all the problems caused by automations. Basically, the module was loaded from version 16.0.
And, to begin with, it doesn't even pass the tests (those of the original module).
There are no model changes or anything.
Sometimes, it's difficult to offer solutions or fixes.
I understand it's not your fault, but it would be nice if someone could share a FAQ with common errors so we can make the necessary fixes and make this task easier.
Thanks.

pedrobaeza:
Please follow the migration guide:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

PR #764, opened 2025-04-14
17.0_MIG_sale_analytic_tag

iTecan:
Migrated from version 16.0.


iTecan:
The module is basically the same, except that the 16.0 mark has been changed to 17.0 because it works correctly and nothing has needed to be modified.
Any guidance?

pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

AlvaroRM11:
Once you have preserved the commit history, don't forget to run the pre-commit checks.

luisDIXMIT:
Hi @iTecan , how is it going with this PR? Need any help?

PR #678, opened 2024-08-12
[16.0][FIX] product_analytic: analytic distribution compute

ljsalvatierra-factorlibre:
When the inverse product compute method is triggered, the current analytic distribution is removed, for example, if you modify the default distribution for an invoice line and duplicate the invoice.

PR #768, opened 2025-04-15
[16.0][REF] product_analytic: use _get_distribution

jbaudoux:
The analytic distribution model serves to provide the analytic account.
Rely on this to collect the product analytic.

Purpose of the change is to be able to collect the proper distribution without directly rely on the move type. For instance if you make a accrued entry from the PO, the move is created in a miscellaneous journal of type `entry`. We should still be able to apply the right analytic account on the accrual move. 

This should also fix the issue fixed in https://github.com/OCA/account-analytic/pull/678. I added a test for that.

cc @sbidoul @rousseldenis @lmignon @ljsalvatierra-factorlibre


houssine78 (approved):
LGTM

houssine78:
@jbaudoux it seems you've fixed the init

lmignon (approved):
LGTM (Code review only)

luisDIXMIT (approved):
Functionally tested, LGTM!

jbaudoux:
@OCA/accounting-maintainers Can you merge this approved one?

PR #772, opened 2025-04-17
[18.0][MIG] product_analytic : migration to 18.0

timloic:


Mantux11:
Hey. I found out that there is a small issue in `_inverse_product_id` method.

Steps how to find it:
1. Create an Invoice;
2. Select a product on Invoice Lines that has default analytic account(it should be added automatically when product is selected.)
3. After product is added, select more Analytic Accounts on analytic_distribution field.
4. Then press save.

You'll see that only default Analytic Account is added. Any other Analytic Accounts that has been added is deleted from line.

I found out simple fix for now, maybe there is a better solution, but i'll just post it here, you could use it:
```python
    @api.onchange("product_id")
    def _inverse_product_id(self):
        res = super()._inverse_product_id()
        # Compatibility with `pos_analytic_by_config`
        if self.env.context.get("pos_config_id"):
            return res
        for line in self:
            inv_type = line.move_id.move_type
            if line.product_id and inv_type and inv_type != "entry" and not line.analytic_distribution:
                ana_accounts = (
                    line.product_id.product_tmpl_id._get_product_analytic_accounts()
                )
                ana_account = ana_accounts[INV_TYPE_MAP[inv_type]]
                line.analytic_distribution = (
                    {ana_account.id: 100} if ana_account else False
                )
        return res
``` 

I just added `and not line.analytic_distribution:` to if statement. And it seems to work like expected.

Maybe, test for this case should also be added.
Thanks.

luisDIXMIT (commented):
I'm testing it on Runboat, and I have an issue. I have two products with different analytic accounts. I created an account move with one product, and the analytic account was set successfully. However, I changed the product to the second one, and the analytic account wasn't updated. Is there something wrong?

timloic:
@luisDIXMIT , @dreispt, @Mantux11  can you review it again :pray: ? It's fixed.

dreispt:
@timloic Thanks, it would have been better to amend your migration commit. Can you squash it plese.
Also, can you look at the tests failing?

ByteMeAsap:
@timloic , can you please make the last few changes?

flotho:
Hi @timloic ,

Thanks for this proposal.
Is there any chance to rebase to trigger CI/CD ?

Regards

Mantux11:
@timloic 
Sorry for late reply. I just can't find what was changed to fix _inverse_product_id. Maybe I'm blind or something :D

Saran440:
@timloic Hi, I think this module should be change widget to analytic account with standard odoo.
What do you think? https://github.com/OCA/account-analytic/pull/838

ypapouin:
Hi @timloic, can you please rebase your PR ? (logs are no more available)

dreispt:
Superseded by https://github.com/OCA/account-analytic/pull/850

PR #759, opened 2025-04-09
[18.0][MIG] purchase_analytic_tag

Saran440:
Standard Migrate

- Cherry-pick code v17 from https://github.com/OCA/account-analytic/pull/715

PR #775, opened 2025-04-24
[18.0][ADD] purchase_request_analytic_tag

Saran440:
This module add analytic tags in purchase request line and send it to purchase order line

Depend On:
- [x] purchase_analytic_tag - https://github.com/OCA/account-analytic/pull/759
- [x] purchase_request


dreispt (commented):
Not blocking, but my preferred approach would be to use out-of-the-box Analytic Distributions.

Saran440:
@dreispt Thank you for your review.

But Analytic distribution is not enough for own project.
and when we use `purchase_request` and `purchase_analytic_tag`
it should send `analytic_tag` (if any) from `purchase_request` to `purchase_order` too

dreispt:
My personal POV is that Analytic Tags is a deprecated feature, replaced by Analytic Distribution, and the future direction should be to rather focus on Analytic Dist. support. 
I don't mean to block any work on Analytic Tags, but I don't thing it is he best place to spend effort on, sorry.

luisDIXMIT:
> My personal POV is that Analytic Tags is a deprecated feature, replaced by Analytic Distribution, and the future direction should be to rather focus on Analytic Dist. support. I don't mean to block any work on Analytic Tags, but I don't thing it is he best place to spend effort on, sorry.

Hi @dreispt  , as far as I understand, analytic distribution models are used with the intention of automatically applying an analytic account to an accounting entry (and subsequently generating an analytic line for that account). However, the purpose of analytic tags is to tag the analytic lines based on the analytic account set on the accounting entry. Therefore, the distribution model determines the analytic account used in the accounting entry, and the tags determine the categorization of the analytic lines based on the analytic account used.

From my point of view, these are not interchangeable features; however, one might question the usefulness of analytic tags — but if they are used, they are necessary, IMHO.
What do you think about this? Could I be missing something?

purchase_request_analytic_tag/models/purchase_request_line.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 Ecosoft Co., Ltd. (http://ecosoft.co.th)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class PurchaseOrderLine(models.Model):
```
luisDIXMIT: class name must be PurchaseRequestLine?

luisDIXMIT (commented):
Tested on runboat and LGTM. Look at my comment!

PR #778, opened 2025-05-06
[18.0][MIG] analytic_hr_department_restriction: Migration to 18.0

victoralmau:
Migration to 18.0

@Tecnativa TT56200


carlos-lopez-tecnativa (commented):
If the user does not have the `Analytic Accounting (without department)` group, I get an error.
I remember that this error did not occur in version 16.

image

Plan without department
image

image

victoralmau:
> If the user does not have the `Analytic Accounting (without department)` group, I get an error. I remember that this error did not occur in version 16.

Fixed.

carlos-lopez-tecnativa (approved):
LGTM

analytic_hr_department_restriction/models/account_analytic_plan.py
```diff
@@ -2,13 +2,23 @@
...
 
 class AccountAnalyticPlan(models.Model):
     _inherit = "account.analytic.plan"
 
-    department_id = fields.Many2one(
-        comodel_name="hr.department",
-        string="Department",
-        domain="['|',('company_id', '=?', company_id),('company_id', '=', False)]",
-    )
+    department_id = fields.Many2one(comodel_name="hr.department", string="Department")
```
pedrobaeza: Why removing the domain?

victoralmau: The company field no longer exists in the plans since v17.

pedrobaeza: Yeah, but this is `hr.department`, which has company, so this should be converted to company dependent or be a many2many.

victoralmau: Ok, change to many2many.

analytic_hr_department_restriction/models/ir_rule.py
```diff
@@ -0,0 +1,59 @@
...
+            and test_condition
+        ):
+            department_f_name = (
+                "department_id"
+                if model_name == "account.analytic.account"
+                else "department_ids"
+            )
+            if user.has_group(group2):
+                extra_domain = [
+                    (department_f_name, "child_of", user.employee_ids.department_id.ids)
```
pedrobaeza: `child_of` works properly with many2many?

victoralmau: It doesn't work correctly, I'm going to follow the way of making the company_dependent field https://github.com/OCA/account-analytic/pull/778#discussion_r2076890490

analytic_hr_department_restriction/migrations/18.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,43 @@
...
+
+from openupgradelib import openupgrade
+
+from odoo.tools.sql import column_exists
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Set the department_id value in the project company."""
+    if not column_exists(env.cr, "account_analytic_plan", "old_department_id"):
```
pedrobaeza, line 12: Why not using openupgradelib method for doing the company dependent transformation?

victoralmau: What would be the method to do this automatically? At https://github.com/OCA/helpdesk/pull/611 it was necessary to do this whole process to get the values right.

pedrobaeza: https://github.com/OCA/openupgradelib/blob/20a96fff6490c379598926ce2d737bb83b3b6af6/openupgradelib/openupgrade.py#L3533

analytic_hr_department_restriction/README.rst
```diff
@@ -0,0 +1,115 @@
...
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+This module allows you to limit plans and analytical accounts by
+employee departments.
+
+Users will only be able to view plans and analytical accounts linked to
+their employee's department or those with no department defined.
+
+Users with the "Invoicing / Billing" permission will be able to view all
```
luisDIXMIT: Please, can you update it to Invoicing / Invoicing? group_account_invoice is now named "Invoicing" https://github.com/odoo/odoo/blob/752ba2927e0b61cd6315e5c780a5292a6e99fe17/addons/account/security/account_security.xml#L60

Same in lines 59 and 64

luisDIXMIT (commented):
Hi @victoralmau , I have a question. A user without the Invoicing / Invoicing permission should only be able to access those analytic accounts/plans that belong to their department. However, a user without this permission can't access anything related to invoicing/accounting, let alone analytic accounting. What am I doing wrong? Thanks.

victoralmau:
> Hi @victoralmau , I have a question. A user without the Invoicing / Invoicing permission should only be able to access those analytic accounts/plans that belong to their department. However, a user without this permission can't access anything related to invoicing/accounting, let alone analytic accounting. What am I doing wrong? Thanks.

This module only limits the analytical plans/accounts that the user can access.

It is possible to list the analytical plans/accounts from different sides, for example, if you have the purchase permission and the “Analytical accounting” permission you can set an analytical distribution on the purchase order lines, and there you are showing the “limited” analytical plans/accounts.

luisDIXMIT (approved):
Thanks you for the explanation @victoralmau I tested it on runboat and LGTM!

pedrobaeza (approved):
/ocabot merge nobump

PR #786, opened 2025-05-22
[16.0][FIX] product_analytic: Use record and not recordset in compute method + remove `depends_context`

rousseldenis:


rousseldenis:
@jbaudoux

product_analytic/models/account_move_line.py
```diff
@@ -1,17 +1,16 @@
...
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
-    @api.depends_context("move_type")
     def _compute_analytic_distribution(self):
         for line in self:
             move_type = self.env.context.get("move_type", line.move_id.move_type)
             super(
-                AccountMoveLine, self.with_context(move_type=move_type)
+                AccountMoveLine, line.with_context(move_type=move_type)
```
jbaudoux, line 14: :O

product_analytic/models/account_move_line.py
```diff
@@ -1,17 +1,16 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import api, models
+from odoo import models
 
 
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
-    @api.depends_context("move_type")
```
jbaudoux, line 10: you sure? Because it's taken from the context inside this method

alexis-via (approved):
I confirm it fixes the bug (I was also impacted for a project I migrated to v16 yesterday !)

pedrobaeza (approved):
/ocabot merge patch

product_analytic/models/account_move_line.py
rousseldenis: @jbaudoux This is because the field is stored. You should not depends context on those ones.

PR #787, opened 2025-05-22
[16.0][TMP] account_analytic_distribution_manual: Temporarily disable test …

carlos-lopez-tecnativa:
…to investigate the cause of the failure.

The error cannot be reproduced locally.


carlos-lopez-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge patch

PR #586, opened 2023-08-23
[16.0][IMP] stock_analytic: more granularity on applicability judgment

AungKoKoLin1997:
Supersede of https://github.com/OCA/account-analytic/pull/543

@qrtl

PR #789, opened 2025-06-02
[18.0][IMP] stock_analytic: more granularity on applicability judgment

dreispt:
Forward port of #586

Before this commit, analytic distribution would be validated only for outgoing moves,
which was not ideal in view of a module such as purchase_stock_analytic.

This commit adds more flexibility to the applicability judgment by adding the stock
picking type field to the analytic applicability model, removing the assumption of
validating outgoing moves only.

### DIFF
--- a/stock_analytic/models/analytic_applicability.py
+++ b/stock_analytic/models/analytic_applicability.py
@@ -11,3 +11,16 @@ class AccountAnalyticApplicability(models.Model):
         selection_add=[("stock_move", "Stock Move")],
         ondelete={"stock_move": "cascade"},
     )
+    stock_picking_type_id = fields.Many2one(
+        "stock.picking.type",
+        string="Operation Type",
+    )
+
+    def _get_score(self, **kwargs):
+        score = super()._get_score(**kwargs)
+        if score >= 0 and self.stock_picking_type_id:
+            if kwargs.get("picking_type") == self.stock_picking_type_id.id:
+                score += 1
+            else:
+                return -1
+        return score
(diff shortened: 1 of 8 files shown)


yostashiro (approved):
Thanks for the port! LGTM.

stock_analytic/models/analytic_applicability.py
```diff
@@ -11,3 +11,16 @@ class AccountAnalyticApplicability(models.Model):
         selection_add=[("stock_move", "Stock Move")],
         ondelete={"stock_move": "cascade"},
     )
+    stock_picking_type_id = fields.Many2one(
+        "stock.picking.type",
+        string="Operation Type",
+    )
+
+    def _get_score(self, **kwargs):
+        score = super()._get_score(**kwargs)
```
FrancoMaxime, line 20: Hi @dreispt, I like adding an ensure_one before each method related to an ensure_one, in the case of the initial method is not longer an ensure_one, all others methods are still protected by the ensure_one  till their update.

What do you think about this?

FrancoMaxime (commented):
LGTM: code review

Hello @dreispt, thanks for the work,
I just have one question.

Great Job!

gbrito (approved):

stock_analytic/models/analytic_applicability.py
dreispt: I have no strong opinion there.
Anyway, I'm not the author, this is a forward port of another person's  work.

PR #792, opened 2025-06-10
[16.0][IMP] analytic_hr_department_restriction: Convert department_id to department_ids

victoralmau:
Convert `department_id` to `department_ids`

Ability to define multiple departments

Please @pedrobaeza and @juancarlosonate-tecnativa can you review it?

@Tecnativa TT56580

PR #797, opened 2025-06-13
[18.0][IMP] analytic_hr_department_restriction: Convert department_id to department_ids

victoralmau:
FWP from 16.0: https://github.com/OCA/account-analytic/pull/792

Convert `department_id` to `department_ids`

Ability to define multiple departments

Please @pedrobaeza and @juancarlosonate-tecnativa can you review it?

@Tecnativa TT56580

### DIFF
--- a/analytic_hr_department_restriction/migrations/18.0.1.1.0/post-migration.py
+++ b/analytic_hr_department_restriction/migrations/18.0.1.1.0/post-migration.py
@@ -1,0 +1,25 @@
+# Copyright 2025 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+from odoo.tools.sql import column_exists
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if column_exists(env.cr, "account_analytic_account", "department_id"):
+        openupgrade.m2o_to_x2m(
+            env.cr,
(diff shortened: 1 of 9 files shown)


analytic_hr_department_restriction/migrations/18.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2025 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.m2o_to_x2m(
```
pedrobaeza: This should be executed only if the field `department_id` is present, as coming directly from 17.0 without this change will fail.

luisDIXMIT (approved):
LGTM!!

pedrobaeza (approved):
/ocabot merge nobump

analytic_hr_department_restriction/migrations/18.0.1.1.0/post-migration.py
luisDIXMIT: This note is not in the original imp for version 16. Should it be?

pedrobaeza: No, on 16 it can't happen, as you come from the version having the field.

PR #798, opened 2025-06-14
[IMP] stock_analytic: more granularity on applicability judgment

dreispt:
**Last mile fixes for https://github.com/OCA/account-analytic/pull/789**
Closes #789

Before this commit, analytic distribution would be validated only for outgoing moves,
which was not ideal in view of a module such as purchase_stock_analytic.

This commit adds more flexibility to the applicability judgment by adding the stock
picking type field to the analytic applicability model, removing the assumption of
validating outgoing moves only.

### DIFF
--- a/stock_analytic/models/stock_move.py
+++ b/stock_analytic/models/stock_move.py
@@ -68,18 +68,29 @@ class StockMove(models.Model):
             res.update({"analytic_distribution": self.analytic_distribution})
         return res
 
+    def _need_validate_distribution(self):
+        """Return moves are made outside the scope of the validation for now, since
+        there could be cases where the necessity cannot be judged solely by the
+        operation type.
+        """
+        self.ensure_one()
+        if self._is_in() and self._is_returned(valued_type="in"):
+            return False
+        elif self._is_out() and self._is_returned(valued_type="out"):
+            return False
+        elif self.company_id.anglo_saxon_accounting and self._is_dropshipped_returned():
+            return False
+        return True
+
     def _action_done(self, cancel_backorder=False):
         for move in self:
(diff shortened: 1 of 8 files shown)


stock_analytic/models/stock_move.py
```diff
@@ -68,18 +68,29 @@ def _prepare_move_line_vals(self, quantity=None, reserved_quant=None):
...
+    def _need_validate_distribution(self):
+        """Return moves are made outside the scope of the validation for now, since
+        there could be cases where the necessity cannot be judged solely by the
+        operation type.
+        """
+        self.ensure_one()
+        if self._is_in() and self._is_returned(valued_type="in"):
+            return False
+        elif self._is_out() and self._is_returned(valued_type="out"):
+            return False
```
guewen, line 80: Hi @dreispt , thanks for adding the applicability, that's a great feature.

I had a hard time understanding why I could confirm some incoming transfers with no analytic distribution when I was expected them to be obligatory. I found out that the implementation for `_is_returned` is pretty naive:

https://github.com/odoo/odoo/blob/c61060bfda17b308efac87bcecbe1d5ae3f3b48d/addons/stock_account/models/stock_move.py#L781-L800

For example with an incoming transfer, it will resolve to (ignoring the transit branch):

```
valued_type == 'in' and self.location_id and self.location_id.usage == 'customer'
```

Which means, for `_is_in()`, it will return True, and the validation will be skipped.

How do you handle this on your side? Do you have an override of `_is_returned` to check if there is a `origin_returned_move_id`?

guewen: Hmm writing this I now realize that customer->stock is probably expected to be a return in most cases, maybe I stumbled upon a very special case.

PR #800, opened 2025-06-18
[17.0][ADD] stock_analytic_rule: New module stock_analytic_rule

BernatObrador:
This module extends the stock and analytic accounting capabilities by automatically creating analytic lines from stock moves based on configurable rules.

Key features:

- Define stock movement rules by source and destination locations.
- Compute analytic line amounts using either:
  - Product list price.
  - Category-based formula: `(avg_price * (avg_weight * qty)) + ((avg_weight * qty) * supplement)`
- Support for positive and negative analytic distributions.
- Handles partial distributions and multi-account combinations.
- Supports return pickings and reversal analytic lines.
- Fully compatible with analytic plans and multidimensional analytic accounting.

The module is useful for organizations needing precise analytic accounting for inventory movements, such as manufacturing, logistics, or services.

cc https://github.com/APSL 4313

@miquelalzanillas @lbarry-apsl @mpascuall @peluko00 @javierobcn @ppyczko please review


mpascuall (approved):
LGTM! But please correct your readme as Stock Analytic Model menu is in invoicing, not in inventory

ppyczko (approved):
LGTM!

peluko00 (approved):
LGTM! Code review and tested in runboat

luisDIXMIT (approved):
Tested on runboat and LGTM!!

BernatObrador:
@pedrobaeza Sorry for pinging you directly. Is this ready to be merged?

pedrobaeza:
Why don't you use and expand `stock_analytic`?

BernatObrador:
@pedrobaeza After further consideration, I realized that relying on *stock_analytic* is not the best option in every case. By not extending it, we avoid the complexity of dealing with the analytic distribution widget. Instead, this module provides the ability to automatically create both positive and negative analytic lines, with a high level of configurability that would be difficult to achieve through *stock_analytic*.

Let me know your thoughts.

pedrobaeza:
The word `model` is confusing, as that term is used for other things inside Odoo. Why don't you call it `stock_analytic_rule` or similar?

BernatObrador:
@pedrobaeza that makes a lot of sense, just changed the module name and everything related to it.

Could you review again please?

Thanks a lot for your time!

PR #812, opened 2025-07-28
[15][MIG] product_analytic_sale: migration to 15.0

PolComas:
This PR backports the core functionality of version 16.0 of the product_analytic_sale module to the 15.0 branch, with the following key changes:

1. In‑place implementation of the compute:
- Since Odoo 15.0 lacks both the analytic.mixin and the analytic_distribution field, we’ve copied and adapted the     _compute_analytic_distribution logic directly into this module.

Now:
    - It retrieves the product’s “income” analytic account via _get_product_analytic_accounts(),
    - Searches for—or creates—a corresponding account.analytic.distribution record with percentage = 100 and the   required tag_id,
    - Assigns that record to the sale.order.line.analytic_distribution field.

2. Differences from v16.0
We have not backported the _prepare_invoice_line override, so analytic distributions will not automatically propagate to invoice lines as they did in v16.

3. Unit tests
Updated tests now verify:

- That analytic_distribution returns an account.analytic.distribution record with the correct account_id, percentage = 100, tag_id, and name.
- That the same distribution record is reused on subsequent lines.
- That it returns False when there is no product_id or only an “expense” account.


product_analytic_sale/models/sale_order_line.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    analytic_distribution = fields.Many2one(
+        "account.analytic.distribution",
+        compute="_compute_analytic_distribution",
+        readonly=True,
+        help="Distribució analítica per defecte definida al producte o categoria",
```
ivilata, line 15: I guess this should be in English.

ivilata (approved):
Looks good to me with the noted very minor exception of the help message in Catalan. Thanks!

etobella (changes requested):
Follow migration guidelines:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

You have lost all history of commits. Also, the migration should have just one commit.

PR #814, opened 2025-07-31
[15][MIG] product_analytic_sale: migration to 15.0

PolComas:
This PR backports the core functionality of version 16.0 of the product_analytic_sale module to the 15.0 branch, with the following key changes:

1. In‑place implementation of the compute:
- Since Odoo 15.0 lacks both the analytic.mixin and the analytic_distribution field, we’ve copied and adapted the _compute_analytic_distribution logic directly into this module.

Now:
- It retrieves the product’s “income” analytic account via _get_product_analytic_accounts(),
- Searches for—or creates—a corresponding account.analytic.distribution record with percentage = 100 and the required tag_id,
- Assigns that record to the sale.order.line.analytic_distribution field.

2. Differences from v16.0
- We have not backported the _prepare_invoice_line override, so analytic distributions will not automatically propagate to invoice lines as they did in v16.

3. Unit tests
Updated tests now verify:
- That analytic_distribution returns an account.analytic.distribution record with the correct account_id, percentage = 100, tag_id, and name.
- That the same distribution record is reused on subsequent lines.
- That it returns False when there is no product_id or only an “expense” account.


PolComas:
Hi @alexis-via could you please take a look at this migration to 15.0 and merge if it looks good? Thanks!

luisDIXMIT (commented):
Helper text in fields must be in English and done the translation.

luisDIXMIT (commented):
I tested the module and it works but, why is the analytic distribution field created in the sale.order.line model but not shown in the view? It should be editable by the user as in v16.

PR #819, opened 2025-08-19
[18.0][FIX] analytic_base_department: Add missing field account_department_id

Puriboonsiri:
**Issue**
When enabling **Task Logs** in *Settings → Project → Time Management*, Odoo raises an error and the feature cannot be activated.

## Steps to Reproduce
1. Go to **Settings**.
2. Open the **Project** tab.
3. Under **Time Management**, enable **Task Logs (Track time spent on projects and tasks)**.

## Support info
- Odoo version: 18.0  
- Modules: `project`, `timesheet`, `project_timesheet_holidays` (if installed)  
- User access: **Settings → Project → Time Management**

### DIFF
--- a/analytic_base_department/__manifest__.py
+++ b/analytic_base_department/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Base Analytic Department Categorization",
     "summary": "Add relationship between Analytic and Department",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.0.1",
     "author": "Camptocamp, Daniel Reis, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Generic Modules/Projects & Services",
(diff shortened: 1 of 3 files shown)


luisDIXMIT (commented):
With this fix, the issue is resolved, but from my point of view the new field does not make sense. The field 'account_department_id' was created in account.analytic.line to relate the department_id field from the account.analytic.account model via the account_id field defined in the same model. In this case, it is pointing directly to the 'department_id' field from timesheet.analysis.report, without taking into account any relation with account.analytic.account.

The problem is that in version 17.0, Odoo modified the view hr_timesheet_report_search in hr_timesheet/report/hr_timesheet_report_view.xml, removing the creation of a new view to establish an inherited view from analytic. This makes it sensitive to modifications in the analytic view, which must be propagated to the timesheet.analysis.report model.

Unfortunately, I don't have an alternative solution to help you. Anyway, I will mark it as Approved, because I have tested it in Runboat, when some people from the community agree with the decision taken.

luisDIXMIT:
By other hand, you should squash commits.

PR #826, opened 2025-08-29
[18.0][ADD] account_analytic

andrel-exo:
This module is based on `purchase_analytics` and adapts the same functionality for account move.

The goal of this module is to ease analytic distribution management on account move. This module add analytic distribution on account move.

If all lines of the account move have the same analytic distribution, the analytic distribution on the account move is automatically set with this value. If an analytic distribution is set on the account move, all lines of the account move will take this value.


account_analytic/views/account_move_views.xml
```diff
@@ -0,0 +1,25 @@
...
+        <field name="name">account.move.form (account_analytic)</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@id='header_left_group']" position="inside">
+                <field
+                    name="analytic_distribution"
+                    widget="analytic_distribution"
+                    groups="analytic.group_analytic_accounting"
+                    readonly="state in ('done', 'cancel')"
```
luisDIXMIT, line 13: Selection 'done' doesn't exists in state fields.

luisDIXMIT (approved):
Tested on runboat and LGTM!! Just fix what was commented.

luisDIXMIT (commented):
Test are failing.

flotho (changes requested):
Hi, thanks for this PR
IMHO the module name is not appropriate because it could cause confusion with native odoo even if it's not exactly the same.

Regards

PR #831, opened 2025-09-04
[18.0][FIX] pos_analytic_by_config: restrict analytic models to backend vs PoS

andrel-exo:
In 16.0 the module injected `pos_config_id` into the distribution logic so that analytic rules with a PoS set were only applied to invoices generated from that PoS.

In 18.0 this propagation was missing, so analytic rules with `pos_config_id` also applied on backend invoices.

This PR fixes the regression by overriding `_get_applicable_models` to always add `pos_config_id` from the context into `vals`. As a result:

- PoS invoices apply models whose `pos_config_id` equals the current POS, plus generic models where `pos_config_id` is unset.
- Backend invoices apply only generic models where `pos_config_id` is unset.

This restores the expected behavior and aligns 18.0 with 16.0.

### DIFF
--- a/pos_analytic_by_config/models/analytic_distribution_model.py
+++ b/pos_analytic_by_config/models/analytic_distribution_model.py
@@ -12,10 +12,8 @@ class AccountAnalyticDistributionModel(models.Model):
         help="Select a Point of Sale for which the analytic distribution will be used",
     )
 
-    def _get_distribution(self, vals):
-        vals = vals.copy()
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please squash both commits into one.

luisDIXMIT (approved):
LGTM!!

pedrobaeza (approved):
/ocabot merge patch

PR #832, opened 2025-09-08
[18.0][MIG] account_analytic_distribution_widget_rebalance

BhaveshHeliconia:


luisDIXMIT (commented):
To fix test errors, JS should be migrated to OWL, or at least QUnit tests should be migrated to Hoot. Both are preferred and will ensure that the tests pass.

PR #833, opened 2025-09-13
[16.0][IMP]analytic_mixin_analytic_account and stock_analytic

arnaudlayec:
### analytic_mixin_analytic_account
Make it compatible with other module **account_financial_report** of **OCA/account-financial-reporting**
Just add `compute_sudo=True` to both fields `analytic_account_ids` and `analytic_account_names`, so that the console does not throw the error `account.move.line: inconsistent 'compute_sudo' for computed fields: analytic_account_ids, analytic_account_names`
Indeed this last module also computes a field `analytic_account_ids` with `store=True

### stock_analytic
Add a `_compute_analytic_distribution` method, quite generic, to apply by default the analytic distribution model at `stock.move` creation, like on any other object using analytic (purchase order, invoice, ...).


arnaudlayec:
Sorry for the 2 commits of different modules in the same PR, it was not intentional and I did not find how to easily correct this

PR #835, opened 2025-09-29
[17.0][IMP] account_move_update_analytic: Add the option to update analytics from the list of moves

victoralmau:
Add the option to update analytics from the list of moves

It is now possible to modify the analytics for all invoice lines of the selected moves

Please @pedrobaeza and @eduezerouali-tecnativa can you review it?

@Tecnativa TT58198

### DIFF
--- a/account_move_update_analytic/__init__.py
+++ b/account_move_update_analytic/__init__.py
@@ -1,1 +1,2 @@
+from . import models
 from . import wizards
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):
Tested and working.

pedrobaeza:
Maybe it would be good to track the fields in the move chatter?

eduezerouali-tecnativa (approved):
code review. LGTM

victoralmau:
> Maybe it would be good to track the fields in the move chatter?

It seems that it is not possible to define tracking in json fields.
`Unsupported tracking on field analytic_distribution (type json).`

PR #838, opened 2025-10-02
[18.0][MIG] product_analytic

Saran440:
This PR enhance analytic of product to support analytic distribution
However, analytic distribution is required `company_id` but product.category don't have company. So, This module must depend on `product_category_company` https://github.com/OCA/multi-company/pull/803

Depend On:
- [ ] product_category_company - https://github.com/OCA/multi-company/pull/803

On display analytic of product
Selection_003

On display analytic of product category
Selection_004


ThiagoMForgeFlow:
Thank you for your contribution — I think it’s closer to what Odoo introduces in version 18. However, this proposal still shows a few limitations:

- If there are Analytic Distribution Models that can also be applied to a product in an Account Move (for example, based on the partner, account prefix, or the product itself), they are not concatenated. Only the result from the Analytic Distribution Models is used. In my opinion, these distributions should be concatenated instead.
- This behavior only applies to Account Moves. For instance, if we install the purchase_analytic module, it won’t affect the purchase order lines of this product.
- Regarding the company_id in the product category, perhaps the best solution would be to add a company_dependent field rather than adding a module dependency. That dependency could cause issues in environments where categories are shared across companies.

Considering all this, I think we could plan a refactor of the module for version 18. In the product and category views, we could add the analytic_distribution_model_ids field and make it editable directly from those views. This approach would leverage Odoo’s standard functionality and address the issues mentioned above.

What do you think?

Saran440:
@ThiagoMForgeFlow Thank you for your feedback.

This behavior only applies to Account Moves. For instance, if we install the purchase_analytic module, it won’t affect the purchase order lines of this product.
- Yes. So, we must have `product_analytic_<module> `series module 

I also agree that refactor this module makes sense.
I’ll review the approach using the `analytic_distribution_model_ids ` field as you suggested.

ThiagoMForgeFlow:
> @ThiagoMForgeFlow Thank you for your feedback.
> 
> This behavior only applies to Account Moves. For instance, if we install the purchase_analytic module, it won’t affect the purchase order lines of this product.
> 
>     * Yes. So, we must have `product_analytic_<module> `series module
> 
> 
> I also agree that refactor this module makes sense. I’ll review the approach using the `analytic_distribution_model_ids ` field as you suggested.

Hi @Saran440, 
Have you been able to work on the refactor that we mentioned?

ThiagoMForgeFlow:
Hi @Saran440,
I’ve implemented a proposal for managing the analytic distribution models — what do you think?
https://github.com/OCA/account-analytic/pull/850

Saran440:
@ThiagoMForgeFlow Thank you for your work. I will check.

Close this PR and move to https://github.com/OCA/account-analytic/pull/850

PR #841, opened 2025-10-10
[18.0][FIX] pos_analytic_by_config

andrel-exo:
My first PR #831 fixed the issue where analytic distributions intended only for POS were also being applied in the backend.
However, it introduced a side effect: analytic lines started being created in flows where they shouldn’t (e.g. POS payments and session closing).

The root cause was that the _get_default_search_domain_vals method was missing. Without it, the POS-scoped models were still being matched globally in Odoo 18, since the analytic engine only filters on keys present in vals.

This new PR restores the original logic but also defines _get_default_search_domain_vals with `{"pos_config_id": False}`.


pedrobaeza (commented):
I don't really get the problem, and it would be advisable:

1. Put comments on the code on what and why that 2 overrides.
2. Put a regression test checking what this patch is fixing.

PR #849, opened 2025-11-07
[18.0][IMP] account_analytic_parent: allow search name parent

Saran440:
This PR implement for allow search parent analytic like product category
Selection_019


AaronHForgeFlow (commented):
:+1:

PR #853, opened 2025-11-26
[19.0][MIG] account_analytic_parent

weinni2000:
fixed new _read_group and changed archive method


rven:
See https://github.com/OCA/account-analytic/pull/886

PR #870, opened 2026-02-04
[18.0][FIX] analytic_hr_department_restriction: Avoid error if the user does not have read access to account.analytic.plan

victoralmau:
Avoid error if the user does not have read access to `account.analytic.plan`

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT60709

### DIFF
--- a/analytic_hr_department_restriction/models/account_analytic_plan.py
+++ b/analytic_hr_department_restriction/models/account_analytic_plan.py
@@ -22,6 +22,10 @@ class AccountAnalyticPlan(models.Model):
             "test_analytic_hr_department_restriction"
         )
         if test_condition:
-            project_plan = self.search([("id", "in", project_plan.ids)])
-            other_plans = self.search([("id", "in", other_plans.ids)])
+            # Avoid error if the user does not have read access to account.analytic.plan
+            if self.has_access("read"):
+                project_plan = self.search([("id", "in", project_plan.ids)])
+                other_plans = self.search([("id", "in", other_plans.ids)])
+            else:
+                project_plan = other_plans = self
         return project_plan, other_plans


analytic_hr_department_restriction/models/account_analytic_plan.py
```diff
@@ -21,7 +21,8 @@ def _get_all_plans(self):
         test_condition = not config["test_enable"] or self.env.context.get(
             "test_analytic_hr_department_restriction"
         )
-        if test_condition:
+        # Avoid error if the user does not have read access to account.analytic.plan
+        if test_condition and self.has_access("read"):
```
pedrobaeza: Shouldn't we do then this in sudo for still see applicable restrictions?

pedrobaeza (commented):
Oh, wait, now I'm seeing that the idea of this overwrite is to filter by permissions, so applying sudo is not correct. Then I think you should do it other way for getting 0 records on each if no permissions.

victoralmau:
> Oh, wait, now I'm seeing that the idea of this overwrite is to filter by permissions, so applying sudo is not correct. Then I think you should do it other way for getting 0 records on each if no permissions.

I disagree that we should change the behavior to “not return any plans” because `analytic` in https://github.com/odoo/odoo/blob/ba0b3118bc0a628989c25e2159e3614c76847a0e/addons/analytic/models/analytic_plan.py# L109 obtains them with `sudo()`; this module **ONLY** filters the “plans that correspond” by department with search; that is why I had previously applied the fix (`has_access(“read”)`) instead of `sudo()`.

pedrobaeza:
But not filtering will return all the plans, so the restriction won't apply... or am I wrong?

victoralmau:
Another approach applied, do you think that's right?

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #876, opened 2026-02-20
[18.0][IMP] stock_analytic: add analytic in stock.valuation.layer

Saran440:
This PR add analytic distribution on Valuation
Screenshot 2026-02-20 091509
Screenshot 2026-02-20 091516


stock_analytic/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2026 Ecosoft Co., Ltd. (<http://ecosoft.co.th>)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import fields, models
+
+
+class StockValuationLayer(models.Model):
+    _name = "stock.valuation.layer"
+    _inherit = ["stock.valuation.layer", "analytic.mixin"]
```
yostashiro: I suppose we don't need to inherit `analytic.mixin`?

Saran440: If we don't inherit `analytic.mixin` it can't use widget="analytic_distribution"

Error:
`ValueError: Invalid field 'analytic_precision' on model 'stock.valuation.layer'`

yostashiro: I think we should then consider adding the analytic_precision field instead in the same manner as the mixin field. Inheriting mixin will add unnecessary overhead.

Saran440: Fixed, Thank you for your suggestion.

yostashiro (approved):
Code review and functional test. 👍

AaronHForgeFlow (approved):
Code LGTM

PR #883, opened 2026-03-12
[18.0][IMP] account_move_analytic_link: adjust account move button widths

LauraCForgeFlow:
Adjusted the column width of the analytic link button in the list view so it no longer takes up unnecessary space. This solution is based on existing Odoo core patterns ([reference](https://github.com/odoo/odoo/blob/96de54ff83965a0e2c0ec10b8ca982483071b149/addons/mrp_subcontracting/views/stock_picking_views.xml#L20)).

Without this PR:

image

With this PR:

image


JordiMForgeFlow (approved):

AndreuOForgeFlow (approved):

AaronHForgeFlow (approved):
👍

PR #887, opened 2026-03-20
[16.0][FIX] purchase_analytic drop column project_id for migration to 18 issue

sergiocorato:
In this version this module renamed the colum `project_id` but do not dropped it, so in the 18.0 this is reused, in a migration without cleaning, from the core module `project_purchase`, creating issue in some cases.

This PR fix this issue, dropping the column if exists.


purchase_analytic/migrations/16.0.2.1.1/post-migration.py
```diff
@@ -0,0 +1,13 @@
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.column_exists(env.cr, "purchase_order", "project_id"):
+        # drop `project_id` column as it will be re-used in v. 18.0 from the core
```
hbrunn:
```suggestion
        # drop `project_id` column from v15 as it will be re-used in v. 18.0 from the core
```

hbrunn (changes requested):
please prefix your pr title with the version

and I suggest to do

```python
openupgrade.rename_columns(env.cr, {'purchase_order': [('project_id', None)]}) 
```

instead to make this reversible.

There's also the danger that this is going to interfere with custom modules that add a column called `project_id`, so as a further safeguard, you can check if there's an `ir.model.fields` entry for the column, and not move it away in that case

purchase_analytic/migrations/16.0.2.1.1/post-migration.py
sergiocorato: Thanks, I've done the changes.

hbrunn (approved):
thanks for the quick followup!

PR #889, opened 2026-03-24
[18.0][ADD] account_analytic_plan_applicability_product

hbrunn:
This module allows to restrict the applicability of analytic plans to specific products instead of a product category.

Configuration
=============

To configure this module, you need to:

- Go to an analytic plan
- Select the product(s) you want to use to restrict the applicability of this plan

Usage
=====

To use this module, you need to:

- Go to an invoice/vendor bill/accounting item and select a product
- Observe the selected product influences the available analytic plans if configured accordingly

Migration of unmerged https://github.com/OCA/account-analytic/pull/556


hbrunn:
tests are fixed in https://github.com/OCA/oca-ci/pull/123

account_analytic_plan_applicability_product/readme/CONFIGURE.md
```diff
@@ -0,0 +1,5 @@
+To configure this module, you need to:
+
+1.  Go to an analytic plan
+2.  Select the product(s) you want to use to restrict the applicability
+    of this plan
```
yostashiro, line 5: It might be good to add this explanation?

```suggestion
    of this plan

To avoid ambiguous matches, configure product-specific rules together with their corresponding product category, and split applicability lines by category when needed. The module doesn't give higher priority to product-specific rules (with no category) over category-specific rules.
```

yostashiro (approved):
Code review. LGTM.

account_analytic_plan_applicability_product/readme/CONFIGURE.md
hbrunn: oh, great point. how about making a product match +1.5? Then if all other things equal, a product only match wins against a category only match

yostashiro: It might work but I don't feel very comfortable with universally putting more weight on products than other criteria.

What I think might be good is to introduce a tie-break rule (e.g. if product-specific record and category specific record have the same score, prioritize the product-specific one). I'm not sure how involved it would be though.

hbrunn: yeah tie breaking is a problem the way it's implemented in analytic. A sequence field would be nice, then people could order stuff they way they want same weight lines to be resolved, but that would be another module.

So I went with a somewhat more elaborate version of your original suggestion. Tables don't look great with our markdown2html conversion, but not much to do about that here

PR #891, opened 2026-04-06
[16.0][FIX] analytic_hr_department_restriction: auto-populate department on analytic account create

giarve:
When creating a new project, Odoo creates a new analytic account without department_ids. The ir.rule restriction then blocks the operation because the empty department_ids doesn't match the user's department domain.

Auto-populate department_ids from the user's employee department(s) when creating analytic accounts without explicit department_ids, so the ir.rule passes naturally without weakening the security restriction.

### DIFF
--- a/analytic_hr_department_restriction/models/account_analytic_account.py
+++ b/analytic_hr_department_restriction/models/account_analytic_account.py
@@ -1,7 +1,7 @@
 # Copyright 2025 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import Command, api, fields, models
 
 
 class AccountAnalyticAccount(models.Model):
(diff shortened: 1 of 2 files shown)


victoralmau (commented):
I don't think this change is necessary. I mean, is it a good idea to define departments based on the user creating the analytics account? In my opinion, no—this will actually cause much more confusion than it will provide benefits.

The right approach is to allow only users with elevated permissions to create analytics accounts and to let them define the departments as they see fit.

giarve:
The issue I found here is when I give analytic access to the Projects app user as project administrator, and they try to create a new project (with a new analytic account), the module is preventing them from creating this analytic account (which happens automatically).

EDIT: the user does not need access to all departments as they are not elevated permissions, only to R&D department, for instance.

giarve:
The main point here is that as project user you should not need to have analytic account permissions to create projects just as it happens without this module installed. Without this module installed you will not find any error whatsoever and the analytic will keep working.

How to solve this properly without adding departments from the user, we can discuss.

PR #901, opened 2026-04-25
[FIX] account_analytic_tag: fix broken xpaths for Odoo 19 view structure

digiter-tw:
## Summary

Two breaking view structure changes in Odoo 19 prevent `account_analytic_tag` from installing:

### 1. `invoice_line_ids` list element renamed
The `<list>` inside `invoice_line_ids` was given `name="journal_items"` in Odoo 19.
The xpath must match `list[@name='journal_items']` instead of bare `list`.

### 2. `line_ids` renamed to `journal_line_ids`
The Journal Items tab field was renamed from `line_ids` to `journal_line_ids` in Odoo 19.

## Changes
- `account_analytic_tag/views/account_move_views.xml`: update two xpaths to match Odoo 19 view structure

## Test
Verified on Odoo 19.0 Community — module installs and `analytic_tag_ids` column appears correctly in both Invoice Lines and Journal Items tabs.

### DIFF
--- a/account_analytic_tag/views/account_move_views.xml
+++ b/account_analytic_tag/views/account_move_views.xml
@@ -5,8 +5,9 @@
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_move_form" />
         <field name="arch" type="xml">
+            <!-- invoice_line_ids list was given name="journal_items" in Odoo 19 -->
             <xpath
-                expr="//field[@name='invoice_line_ids']/list/field[@name='analytic_distribution']"
+                expr="//field[@name='invoice_line_ids']/list[@name='journal_items']/field[@name='analytic_distribution']"
                 position="after"
             >
                 <field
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't get the changes. They were correct AFAIK.

cristina-hidalgo-tecnativa (changes requested):
Hi @digiter-tw, thanks for the contribution!

After checking the Odoo 19 source code, the original xpaths are correct:

- `invoice_line_ids`: XPath `list/field[...`]  matches any `<list>` regardless of its name attribute, so no fix is needed here. [source](https://github.com/odoo/odoo/blob/681610c002a310f1c73fc2e5bec8d3dae27bc4a7/addons/account/views/account_move_views.xml#L1400)
- `line_ids`: The view still uses `line_ids`, not `journal_line_ids`. That field exists in the model but is marked as [deprecated](https://github.com/odoo/odoo/blob/681610c002a310f1c73fc2e5bec8d3dae27bc4a7/addons/account/models/account_move.py#L189) and unused in views. Changing the xpath actually breaks the installation — confirmed on runboat:

```
Element '<xpath expr="//field[@name=&#39;journal_line_ids&#39;]/list/field[@name=&#39;analytic_distribution&#39;]">' cannot be located in parent view

View error context:
{'file': '/mnt/data/odoo-addons-dir/account_analytic_tag/views/account_move_views.xml',
 'line': 1,
 'name': 'account.move.tag_ids.form.inherit',
 'view': ir.ui.view(861,),
 'view.model': 'account.move',
 'view.parent': ir.ui.view(709,),
 'xmlid': 'account_move_analytic_tag_ids_form'}
```

The changes are not necessary and the current code in 19 version is correct as-is.

PR #907, opened 2026-05-01
[18.0][IMP] stock_analytic: add option to assign analytic on valuation account

yostashiro:
Add a configurable setting (Stock Analytic On Valuation) to allow analytic distribution to be assigned to journal items for the stock valuation account, which is excluded by default.

@qrtl QT6699


SanamiK (approved):
Functional review. LGTM.

AungKoKoLin1997 (approved):
Code Review: 👍

BhaveshHeliconia (approved):
LGTM!

Saran440:
@yostashiro Thanks for your work.
However, in Odoo 18 the core behavior has changed with the introduction of analytic distribution from the sale order line:
https://github.com/odoo/odoo/pull/268031

With this change, the analytic distribution defined on the sale order line is propagated to all related stock valuation move lines.

So, we should introduce a configuration option in this module to support different behaviors:
1. Apply analytic distribution to all stock valuation lines (current standard Odoo behavior).
2. Apply analytic distribution only to COGS lines (similar to the previous core behavior).

What do you think?


Standard without install stock_analytic
selection_340
selection_341

PR #914, opened 2026-05-07
[17.0][FIX] account_move_update_analytic: strip unchanged integrity fields before hash check

cristina-hidalgo-tecnativa:
When `restrict_mode_hash_table` is active, updating `analytic_distribution` on a posted invoice triggers `_sync_dynamic_line` to rewrite the tax line with its full key dict — always including `account_id` even if unchanged. The OCB hash check raises a false positive:

```
odoo.exceptions.UserError: You cannot edit the following fields: Account, Label, Partner.
The following entries are already hashed: INV/2026/XXXXX
```

This fix strips integrity-hash fields from `vals` when their value has not actually changed, so the check only fires on real modifications.

**Note**: This issue only occurs when the tax has the "Include in analytic" option enabled (`analytic` field on `account.tax`), as this makes `analytic_distribution` part of the tax key in `_compute_all_tax`, triggering the `_sync_dynamic_line` rewrite.

@Tecnativa TT61686
@christian-ramos-tecnativa @pedrobaeza

### DIFF
--- a/account_move_update_analytic/models/account_move_line.py
+++ b/account_move_update_analytic/models/account_move_line.py
@@ -1,0 +1,30 @@
+from odoo import models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    def write(self, vals):
+        if not vals:
+            return True
+        # The Odoo core hash check blocks writes where integrity fields appear in vals,
+        # even if unchanged. Updating analytic_distribution causes
+        # _sync_dynamic_line to include account_id in the tax line write,
+        # triggering a false positive. Strip unchanged integrity fields from
+        # vals so the check only fires on real modifications.
+        hashed = self.filtered(lambda line: line.move_id.inalterable_hash)
+        if self.env.context.get("update_analytic") and hashed:
+            integrity_fields = set(self._get_integrity_hash_fields()).union(
+                {"inalterable_hash", "secure_sequence_number"}
+            )
+            unchanged = {
+                fname
+                for fname in integrity_fields & set(vals)
+                if not any(
+                    self.env["account.move"]._field_will_change(line, vals, fname)
+                    for line in hashed
+                )
+            }
+            if unchanged:
+                vals = {k: v for k, v in vals.items() if k not in unchanged}
+        return super().write(vals)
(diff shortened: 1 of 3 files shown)


account_move_update_analytic/models/account_move_line.py
```diff
@@ -0,0 +1,30 @@
+from odoo import models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    def write(self, vals):
+        if not vals:
+            return True
+        # The OCB hash check blocks writes where integrity fields appear in vals,
```
pedrobaeza: OCB?

pedrobaeza (approved):
In the steps to reproduce, you have to mention that the tax has the "Include in analytic" flag activated.

christian-ramos-tecnativa (approved):
Thanks can you FWP? The issue is a bit different in 18.0, there is no error but the analytic change is not reflected in the tax line

account_move_update_analytic/models/account_move_line.py
cristina-hidalgo-tecnativa: Odoo core, missclick. Fixed!

remi-filament (approved):
Thanks @cristina-hidalgo-tecnativa 

I am concerned that updating the write on widely used account.move.line for removing fields from vals if unchanged and part of integrity_fields may hide errors elsewhere. 
What is your opinion on this ?

pedrobaeza:
I see the scope very limited, because at the end, this is only preventing to rewrite a value that is the same, and only for that circumstances.

Anyway, and extra safeguard may be to pass a specific context when doing the analytic update, and only react on that case. Do you understand what I mean, @cristina-hidalgo-tecnativa ?

cristina-hidalgo-tecnativa:
> I see the scope very limited, because at the end, this is only preventing to rewrite a value that is the same, and only for that circumstances.
> 
> Anyway, and extra safeguard may be to pass a specific context when doing the analytic update, and only react on that case. Do you understand what I mean, @cristina-hidalgo-tecnativa ?

Do you maybe mean passing a specific context flag from the wizard, and only stripping the unchanged integrity fields when that flag is present?

pedrobaeza:
That's it!

christian-ramos-tecnativa (approved):
Tested again and working fine!

PR #916, opened 2026-05-11
[18.0] ADD analytic_cost_distribution

eLBati:


QuiJoQuim:
Hi,

I’m following this PR with a lot of interest because we have been working on a related full-costing approach around analytic plans, and the timing is surprisingly similar :)

First of all, I really like several aspects of your proposal, especially:

* the simplicity of the analytic tree configuration,
* and the distinction between different analytic line types to drive redistribution behaviour.

We have been exploring similar problems and I think this PR is moving in a very interesting direction.

Our approach partially overlaps, but the conceptual model is a bit different:

* we try to preserve direct costing completely untouched and add a separate full-costing layer instead of replacing the original analytic distribution. This allows users to keep both direct-costing and full-costing views side by side.
* we also allow “reproductive” (support/overhead) plans inside productive analytic trees, computing overhead propagation through the hierarchy itself. A typical example would be commercial or support work attached to a specific business area.
* our core approach does not depend on timesheets because not all companies use them consistently. Instead, we have been considering optional glue modules for project- or HR-related integrations.

So from my point of view the approaches may actually be complementary rather than competing.

I was wondering whether you would be open to discussing possible convergence points or collaboration before things become too fixed architecturally. Maybe some concepts should remain separate, but perhaps we could share infrastructure or align some parts of the model.

In any case, I think this is a very valuable contribution.

eLBati:
Hi @QuiJoQuim, thanks so much for your review. We'll discuss it internally and get back to you

PR #917, opened 2026-05-12
[18.0][FIX] account_move_update_analytic: update tax lines analytic on posted moves

cristina-hidalgo-tecnativa:
When using the wizard on a posted invoice, two issues were fixed:

- **Tax lines not updated:** `_sync_tax_lines` skips posted moves, so tax lines with "Include in Analytic Cost" were never updated. The fix updates them explicitly alongside the product line.

- **Wrong tax lines updated:** when multiple products share the same analytic tax, all their tax lines were updated instead of only the edited product's. The fix uses each tax line's `analytic_distribution` to identify which product it belongs to (Odoo sets it from the product line's distribution via `_prepare_base_line_tax_repartition_grouping_key`).

FWP [#914 ](https://github.com/OCA/account-analytic/pull/914)

@Tecnativa TT61686
@pedrobaeza @christian-ramos-tecnativa

### DIFF
--- a/account_move_update_analytic/tests/test_account_move_update_analytic.py
+++ b/account_move_update_analytic/tests/test_account_move_update_analytic.py
@@ -1,0 +1,134 @@
+# Copyright 2026 Tecnativa - Christian Ramos
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+from odoo.fields import Command
+from odoo.tests import common, tagged
+
+
+@tagged("post_install", "-at_install")
+class TestAccountAnalyticTagBase(common.TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(
+            context=dict(
+                cls.env.context,
+                mail_create_nolog=True,
+                mail_create_nosubscribe=True,
+                mail_notrack=True,
+                no_reset_password=True,
+                tracking_disable=True,
+            )
+        )
+        # ==== For Accounting ====
+        cls.default_plan = cls.env["account.analytic.plan"].create({"name": "Default"})
+        cls.analytic_account_a = cls.env["account.analytic.account"].create(
+            {
+                "name": "analytic_account_a",
+                "plan_id": cls.default_plan.id,
+            }
+        )
+        cls.analytic_account_b = cls.env["account.analytic.account"].create(
+            {
+                "name": "analytic_account_b",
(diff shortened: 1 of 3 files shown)


account_move_update_analytic/wizards/account_move_update_analytic.py
```diff
@@ -63,7 +63,30 @@ def update_analytic_lines(self):
...
+                and self.line_id.parent_state == "posted"
+            ):
+                # _sync_tax_lines skips posted moves (state != 'draft' guard in core),
+                # so tax lines with 'Include in Analytic Cost' must be updated
+                # explicitly.
+                move = self.line_id.move_id
+                extra_lines = move.line_ids.filtered(
+                    lambda line: line.display_type == "tax"
+                    and line.tax_line_id.analytic
+                )
```
christian-ramos-tecnativa: You should add a more specific filter, just for the lines linked to the edited one, otherwise it will modify tax lines from other products:

Image

christian-ramos-tecnativa (changes requested):

account_move_update_analytic/wizards/account_move_update_analytic.py
cristina-hidalgo-tecnativa: Done! Added a more specific filter to restrict the update to tax lines whose originating tax is actually applied to the edited product line.

Shide (approved):
Approved since this is a FWP, but not tested at all, just code review.
Would be great to have test to check this case

account_move_update_analytic/tests/test_account_move_update_analytic.py
```diff
@@ -0,0 +1,134 @@
...
+                    ),
+                    (
+                        0,
+                        0,
+                        {
+                            "product_id": cls.product_b.id,
+                            "quantity": 2.0,
+                            "price_unit": 100.0,
+                            "analytic_distribution": {
+                                cls.analytic_account_a.id: 100,
```
eduezerouali-tecnativa, line 88: Please use Command here too. Everything else LGTM

carlos-lopez-tecnativa (approved):
LGTM.

@eduezerouali-tecnativa, I applied your suggestion. Please update your review.

Ping @pedrobaeza.

eduezerouali-tecnativa (approved):

pedrobaeza (approved):
There's a lot of new code, but it seems to not be avoidable with the new cases and all of the split thing.

/ocabot merge patch

PR #923, opened 2026-05-22
[12.0][FIX] account_analytic_default_account: skip catch-all default when searching by account

MiquelRForgeFlow:
When account_id is provided to account_get (move line creation context), a record with no conditions set scores 0 and wins only by default (best_index starts at -1). This causes any unconditional analytic default to be applied to every accounting entry without a specific rule, such as inventory adjustment lines. Return empty when account_id is given and the best match has no conditions set.


AaronHForgeFlow (approved):
👍

PR #930, opened 2026-06-08
[19.0][MIG] stock_warehouse_analytic

glitchov:
Migration from 14.0


FrancoMaxime (approved):
LGTM: code review

BhaveshHeliconia (approved):
Functional review LGTM!

stock_warehouse_analytic/models/stock_warehouse.py
```diff
@@ -1,16 +1,9 @@
...
 class StockWarehouse(models.Model):
-
-    _inherit = "stock.warehouse"
-
-    account_analytic_id = fields.Many2one(
-        comodel_name="account.analytic.account",
-    )
-    account_analytic_tag_ids = fields.Many2many(
-        comodel_name="account.analytic.tag",
-    )
```
FrancoMaxime, line 16: Could you make a migration script because the name is change to analytic_distribution

FrancoMaxime (changes requested):

stock_warehouse_analytic/models/stock_warehouse.py
glitchov: Yes but it's not universal so it's preferable to do it on custom side.

PR #932, opened 2026-06-08
[ADD] stock_request_analytic_tag: add new module

CristianoMafraJunior:
@Escodoo SO571-53
Related #2497

This module adds support for analytic tags on stock requests. Analytic tags can be used to classify and track stock requests by different cost centers, departments, or projects. The tags are automatically propagated to the related stock moves.

cc @victoralmau


stock_request_analytic_tag/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2026 - TODAY, Cristiano Mafra Junior <cristiano.mafra@escodoo.com.br>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "Stock Request Analytic Tag",
+    "version": "16.0.1.0.0",
+    "category": "Accounting & Finance",
+    "website": "https://github.com/OCA/account-analytic",
+    "author": "Escodoo, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["stock_request", "account_analytic_tag"],
```
victoralmau:
```suggestion
    "depends": ["stock_request_analytic", "account_analytic_tag"],
```

stock_request_analytic_tag/views/stock_request_views.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_stock_request_form" model="ir.ui.view">
+        <field name="name">stock.request.form</field>
+        <field name="model">stock.request</field>
+        <field name="inherit_id" ref="stock_request.view_stock_request_form" />
```
victoralmau:
```suggestion
        <field name="inherit_id" ref="stock_request_analytic.view_stock_request_form" />
```

stock_request_analytic_tag/views/stock_request_views.xml
```diff
@@ -0,0 +1,17 @@
...
+        <field name="arch" type="xml">
+            <field name="procurement_group_id" position="after">
```
victoralmau:
```suggestion
            <field name="analytic_account_id" position="after">
```

stock_request_analytic_tag/models/stock_request.py
```diff
@@ -0,0 +1,20 @@
...
+        comodel_name="account.analytic.tag",
+        string="Analytic Tags",
+        domain="['|', ('company_id', '=', False), ('company_id', '=', company_id)]",
+    )
+
+    def _prepare_move_line_vals(self):
+        vals = super()._prepare_move_line_vals()
+        if self.analytic_tag_ids:
+            vals.update(analytic_tag_ids=[Command.set(self.analytic_tag_ids.ids)])
+        return vals
```
victoralmau:
```suggestion
```

This isn't correct; it should look something like https://github.com/OCA/stock-logistics-warehouse/blob/15.0/stock_request_analytic/models/stock_rule.py#L10C5-L10C32

It would also be a good idea to add a test to verify that this works correctly, perhaps https://github.com/OCA/stock-logistics-warehouse/blob/0156a195936ae8b4c6f3ea9f008f257c9155fb64/stock_request_analytic/tests/test_stock_request_analytic.py#L165 might help with an example

victoralmau (changes requested):
This module must depend on `stock_request_analytic`: https://github.com/OCA/stock-logistics-warehouse/pull/2497

stock_request_analytic_tag/models/stock_move.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    analytic_tag_ids = fields.Many2many(
+        comodel_name="account.analytic.tag",
+        string="Analytic Tags",
+        domain="['|', ('company_id', '=', False), ('company_id', '=', company_id)]",
+    )
```
victoralmau, line 14:
```suggestion
```
This module should not add this to this model; if there is a process that creates analytic items and requires the analytical tag, that specific process should be modified to define the analytical tags there (https://github.com/OCA/account-analytic/blob/16.0/stock_analytic/models/stock_move.py por example).

victoralmau (changes requested):

PR #933, opened 2026-06-09
[17.0][FIX] account_move_update_analytic: Add force_delete for journal items merge

christian-ramos-tecnativa:
When performing an analytic update is posible that during the sync_lines if the analytic distribution and tax is the same it will try to merge the journal items. Raising the error:
```
Invalid Operation
You can't delete a posted journal item. Don’t play games with your accounting records; reset the journal entry to draft before deleting it.
```
This should be a safe delete, similar to how it creates new items when analytic distribution differs

@Tecnativa TT62858
@pedrobaeza

### DIFF
--- a/account_move_update_analytic/wizards/account_move_update_analytic.py
+++ b/account_move_update_analytic/wizards/account_move_update_analytic.py
@@ -68,6 +68,6 @@ class AccountMoveUpdateAnalytic(models.TransientModel):
             .browse(self.env.context.get("active_ids"))
             .invoice_line_ids
         )
-        lines.with_context(update_analytic=True).write(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But we shouldn't merge the items, isn't it?

christian-ramos-tecnativa:
It's done upstream after the write, right now if there is 1 journal item with same tax for 2 lines and the analytic is changed for one of them a new journal item is created to hold the tax and new analytic. The amount remains the same on reports etc. Im assuming that if creating a new journal item its ok, merging should also be correct. Tried with a hashed invoice and works fine.

pedrobaeza (approved):
OK then.

eduezerouali-tecnativa (approved):

PR #943, opened 2026-07-08
[18.0][FIX] account_analytic_distribution_manual: avoid save race in widget

MiquelRForgeFlow:
Send analytic_distribution and manual_distribution_id in a single record.update() when the popup closes. With two sequential updates, a form save processed between them re-dirtied the record right after saving, so the form never reached the saved state and the tour timed out waiting for it (flaky CI failure).

Error detected in https://github.com/OCA/account-analytic/actions/runs/28935792037/job/85845684951.


AaronHForgeFlow (approved):
👍

carlos-lopez-tecnativa (approved):
LGTM, thanks

PR #948, opened 2026-07-27
[18.0][ADD] account_analytic_maintenance: Add analytic distribution on maintenance equipment

Ricardo-MC:
This module stores an analytic distribution on maintenance equipment so it can be reused by other maintenance integrations

PR #950, opened 2026-07-27
[18.0][ADD] account_analytic_maintenance_stock: Propagate analytic distribution to maintenance stock moves

Ricardo-MC:
When stock moves are linked to maintenance equipment through the
maintenance_stock module, this module propagates the analytic
distribution configured on the equipment to the stock move.

The implementation is intentionally limited to stock moves related to
maintenance equipment and only fills the analytic distribution when it
has not already been set, preserving values provided by other modules.

Currently maintenance_stock does not provide a dedicated hook to
initialize stock move values, so this module overrides create() with a
narrowly scoped implementation.


celm1990 (changes requested):
LGTM, but couldn't this be assigned before the record is created?

This would avoid triggering a write immediately after create and optimize the flow.

For example, you could retrieve the `default_maintenance_request_id` from the context in default_get or use a similar approach.

celm1990 (changes requested):
Please add some tests as well.

account_analytic_maintenance_stock/tests/test_stock_move.py
```diff
@@ -0,0 +1,122 @@
...
+            {
+                "picking_type_id": self.maintenance_warehouse.cons_type_id.id,
+                "location_id": self.maintenance_warehouse.lot_stock_id.id,
+                "location_dest_id": self.maintenance_warehouse.wh_cons_loc_id.id,
+                "maintenance_request_id": self.request.id,
+                "move_ids": [
+                    (
+                        0,
+                        0,
+                        {
```
celm1990:
```suggestion
                    Command.create(
                        {
```

celm1990 (changes requested):

account_analytic_maintenance_stock/tests/test_stock_move.py
Ricardo-MC: Fixed, thanks!

celm1990:
Depends on

- [ ] https://github.com/OCA/account-analytic/pull/948

account_analytic_maintenance_stock/models/stock_move.py
```diff
@@ -0,0 +1,16 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if vals.get("analytic_distribution") or not vals.get("picking_id"):
+                continue
+            picking = self.env["stock.picking"].browse(vals["picking_id"])
+            equipment = picking.maintenance_equipment_id
+            if equipment and equipment.analytic_distribution:
+                vals["analytic_distribution"] = equipment.analytic_distribution
+        return super().create(vals_list)
```
celm1990: As mentioned in my previous comment, this implementation is not convincing to me. I see that `analytic_distribution` is computed in the mixin, so you can inherit this function to implement the logic:

https://github.com/odoo/odoo/blob/a47a80fb04d113e3da83e654b667e7335cf63d39/addons/analytic/models/analytic_mixin.py#L17

```suggestion
    @api.depends("picking_id.maintenance_request_id")
    def _compute_analytic_distribution(self):
         res = super()._compute_analytic_distribution(vals_list)
         for move in self:
             if move.picking_id.maintenance_request_id.equipment_id.analytic_distribution
                 move.analytic_distribution = move.picking_id.maintenance_request_id.equipment_id.analytic_distribution
             return res

```

account_analytic_maintenance_stock/tests/test_stock_move.py
```diff
@@ -0,0 +1,125 @@
+from odoo import Command
+from odoo.tests import Form, common
+
+
+class TestStockMove(common.TransactionCase):
```
celm1990: Please inherit from `BaseCommon`.
```suggestion
class TestStockMove(BaseCommon):
```

account_analytic_maintenance_stock/tests/test_stock_move.py
```diff
@@ -0,0 +1,125 @@
...
+                            }
+                        )
+                    )
+                ],
+            }
+        )
+        self.assertEqual(
+            picking.move_ids.analytic_distribution,
+            distribution,
+        )
```
celm1990: I think this test is unnecessary because it does not test anything related to this module. The other two tests already cover the flow.

account_analytic_maintenance_stock/tests/test_stock_move.py
```diff
@@ -0,0 +1,125 @@
...
+                "location_id": location.id,
+                "quantity": 5,
+            }
+        )
+        picking_form = Form(
+            self.env["stock.picking"].with_context(
+                default_maintenance_request_id=self.request.id,
+            )
+        )
+        picking_form.picking_type_id = self.maintenance_warehouse.cons_type_id
```
celm1990: I think you should use the context from the action `action_view_stock_picking_ids`: https://github.com/OCA/maintenance/blob/33ae57cec76a3884a01e9567ef3c8d7386022cff/maintenance_stock/models/maintenance_request.py#L25


```suggestion
        action = self.request.action_view_stock_picking_ids()
        picking_form = Form(
            self.env["stock.picking"].with_context(**action["context"])
        )
```

celm1990 (changes requested):
@Ricardo-MC, please review my comments. I think this is a better approach, and the user will see the analytic distribution immediately instead of only after saving the record.

account_analytic_maintenance_stock/tests/test_stock_move.py
Ricardo-MC: @celm1990
Fixed, Thanks!

PR #952, opened 2026-08-03
[18.0][IMP] sale_analytic_distribution_model_warehouse: recompute lines after warehouse change

arielbarreiros96:
Small change to also recompute lines after warehouse change.


christian-ramos-tecnativa (approved):
LGTM!

sale_analytic_distribution_model_warehouse/tests/test_so_analytic_by_warehouse.py
```diff
@@ -93,3 +93,30 @@ def test_create_sale_order(self):
...
+        order.warehouse_id = self.warehouse2
+        self.assertEqual(
+            order.order_line.analytic_distribution,
+            {str(self.analytic_account2.id): 100.0},
+        )
+        warehouse3 = self.env["stock.warehouse"].create(
+            {"name": "Warehouse 3", "code": "WH3"}
+        )
+        order.warehouse_id = warehouse3
+        self.assertFalse(order.order_line.analytic_distribution)
```
pedrobaeza: This is redundant. As the code has been already tested in the previous lines.

sale_analytic_distribution_model_warehouse/__manifest__.py
```diff
@@ -7,7 +7,7 @@
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/account-analytic",
     "category": "Purchases, Accounting",
-    "version": "18.0.1.0.1",
+    "version": "18.0.1.0.2",
```
pedrobaeza: Don't manually change the version. It will be done on merge by the bot.

pedrobaeza (changes requested):

sale_analytic_distribution_model_warehouse/tests/test_so_analytic_by_warehouse.py
arielbarreiros96: Hello @pedrobaeza thanks for the review, always appreciated. This part specifically tests what happens when`order.warehouse_id = warehouse3`, notice that `warehouse3` is the only warehouse that has not been assigned to a `account.analytic.distribution.model`, therefore I would expect that `order.order_line.analytic_distribution` to be `False` (under the conditions of the test), I checked the previous tests and found no trace of that verification, that possibility is introduced now, after the change on `_compute_analytic_distribution`

pedrobaeza: Uhm, true, thanks for insisting. Indeed it's checking another different thing, so yes, it has value. For avoiding this kind of misunderstanding, you can put a comment line at the starting of each of the "blocks" for guiding reviewers. Example: `# Assign a warehouse absent from models: no analytic distribution`.

And this raises another question: should the analytic distribution be reset if not warehouse matches, or to keep the previous one? First of all, other models without warehouse may match. What do you think?

arielbarreiros96: Thanks for the comments, I'm adding the note to the test for clarification, actually I'll start doing so as common practice 👌🏻.

In this case if the warehouse_id changes de distribution for each line gets recomputed from scratch (notice the `line.analytic_distribution = False` before calling `super(SaleOrderLine, line)._compute_analytic_distribution()`), so this do not interfere with other rules, so for example if a distribution model based on `partner_id` adds X or Y distribution this is still respected.

I'm proposing this becuse to cover the following scenario: a user starting a sale order with let's say 20 lines, and only noticed the `warehouse_id` was incorrect when reached line 15, it changed the `warehouse_id`, and added the next 5 lines; in that scenario the SO gets 15 SOL with de distribution of WH0 and 5 with the one corresponding to WH1, which is normally not the intended behaviour. 

I do understand that this sacrifices manual changes to the distribution if any of these change `"order_id.partner_id", "product_id", "order_id.warehouse_id"`, but in that case my question is, what should we prioritize, manual edits or enforcement of the analytic distribution models? 😅 It is indeed an interesting question. I did added a note to the readme, just in case 😅

pedrobaeza: Uhm, but think in the contrary case: you have assigned manually 10 lines the analytic distributions, and then you find that the warehouse is incorrect. You change it, but you don't have any distribution model. You lose everything. On standard, I suppose changing the product resets it, but that's a local change in one line and it's reasonable, not this case.

I think the analytic distribution should be changed for the warehouse change only if there's a match with a distribution model by warehouse.

arielbarreiros96: > Uhm, but think in the contrary case: you have assigned manually 10 lines the analytic distributions, and then you find that the warehouse is incorrect. You change it, but you don't have any distribution model. You lose everything. On standard, I suppose changing the product resets it, but that's a local change in one line and it's reasonable, not this case.
> 
> I think the analytic distribution should be changed for the warehouse change only if there's a match with a distribution model by warehouse.

Let me see if I understood correctly. It would be basically override everything if there is a mathching distribution model (including wiping all manual changes), but leave everything as it is (including manual changes) if there is no matching distribution? I mean for me it makes sense, if there is no matching distribution there is no need to wipe anything

pedrobaeza: Yes, almost. I would restrict to "if there is a matching distribution model with warehouse", as if the matching distribution model doesn't have warehouse, then it wouldn't apply either in normal cases (as they are just triggered when the product changes).

arielbarreiros96: > Yes, almost. I would restrict to "if there is a matching distribution model with warehouse", as if the matching distribution model doesn't have warehouse, then it wouldn't apply either in normal cases (as they are just triggered when the product changes).

Makes sense, I'll push again with those changes, and since this particular commit is included on the migration to v19, once we agree here I'll cherry-pick it on the v19 migration as well 😉

arielbarreiros96: @pedrobaeza it's done, I tuned the tests a bit for the sake of reclycing 😅, but they essentialy remain the same. Plus they now cover the 2 use cases we've discussed so far 👌🏻

pedrobaeza (approved):
Excellent! Thanks for the great dialogue and collaboration.

/ocabot merge patch

PR #958, opened 2026-08-11
[18.0][IMP] account_analytic_maintenance_purchase_request: propagate analytic distribution

Ricardo-MC:
Propagate the analytic distribution from the maintenance equipment to purchase orders linked to maintenance requests.


account_analytic_maintenance_purchase_request/models/purchase_order.py
```diff
@@ -0,0 +1,20 @@
...
+                continue
+            distribution = request.equipment_id.analytic_distribution
+            if distribution:
+                order.analytic_distribution = distribution
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "maintenance_request_ids" in vals:
+            self._propagate_maintenance_analytic_distribution()
+        return res
```
celm1990: I’m not sure this is the best way to implement this.

The user should save the record first before seeing the analytic distribution.

Additionally, purchase.order does not have an analytic_distribution field; it exists at the purchase.order.line level. Only when the purchase_analytic module is installed is this field added to purchase.order. However, this module is not an explicit dependency, so please follow the standard Odoo behavior and handle the analytic distribution at the purchase.order.line level.

I tested this from a maintenance request: I selected or created a purchase order, and I expected the lines to take the analytic distribution only from this specific context. However, the field remains empty, so it is not working as expected.

Image

account_analytic_maintenance_purchase_request/tests/test_purchase_order.py
```diff
@@ -0,0 +1,73 @@
...
+            {
+                "maintenance_request_ids": [
+                    Command.link(self.request.id),
+                ],
+            }
+        )
+        self.assertEqual(
+            purchase.analytic_distribution,
+            distribution,
+        )
```
celm1990: This test is not correct because you cannot set the maintenance request from the purchase order. Or am I missing something here?

I expected that, from the maintenance request, when I create a new purchase order, the purchase order lines would take the analytic distribution from the maintenance equipment.

celm1990 (changes requested):

account_analytic_maintenance_purchase_request/models/purchase_order.py
Ricardo-MC: Thanks for the review. Fixed as suggested. The analytic distribution is now propagated to the purchase order lines before they are created.

Ricardo-MC:
ping @celm1990

celm1990:
Depends on

- [ ] https://github.com/OCA/account-analytic/pull/948

account_analytic_maintenance_purchase_request/models/purchase_order_line.py
```diff
@@ -0,0 +1,23 @@
...
+            request = order.maintenance_request_ids[:1]
+
+            if not request:
+                continue
+
+            distribution = request.equipment_id.analytic_distribution
+            if distribution:
+                vals["analytic_distribution"] = distribution
+
+        return super().create(vals_list)
```
celm1990: This is not working as expected. When I create a purchase order and add lines, then save the record, the analytic distribution is not propagated to the purchase order lines.

Please reimplement the logic.

In purchase_order_ids, add some context to propagate the current maintenance request.

Then, in the compute method of purchase.order or purchase.order.line, inherit the compute function and retrieve the field, similar to the approach suggested in https://github.com/OCA/account-analytic/pull/950#discussion_r3833654371.

account_analytic_maintenance_purchase_request/tests/test_purchase_order_line.py
```diff
@@ -0,0 +1,99 @@
...
+            maintenance_request_ids=[self.request.id],
+        )
+        line = self._create_purchase_line(
+            purchase,
+            analytic_distribution=purchase_distribution,
+        )
+        self.assertEqual(
+            line.analytic_distribution,
+            purchase_distribution,
+        )
```
celm1990: Similar to https://github.com/OCA/account-analytic/pull/950#discussion_r3833682211

celm1990 (changes requested):

Ricardo-MC:
@celm1990
Fixed, Thanks!

PR #960, opened 2026-08-17
[18.0][ADD] account_analytic_payment

rjaraspearhead:
## Module: `account_analytic_payment`

New module that adds an **analytic distribution** on `account.payment`
(inheriting `analytic.mixin`) and propagates it to the accounting entry.

### What it does
- Adds the *Analytic Distribution* field on the payment header
  (widget `analytic_distribution`, under the analytic accounting group).
- On post, the distribution is written to the **counterpart journal item**
  (the receivable/payable line, `destination_account_id`) via an override of
  `_prepare_move_counterpart_lines`.
- If no distribution is set on the payment, no journal item is affected.

### Why the counterpart line
The receivable/payable line is the meaningful one for downstream analysis and
keeps it consistent with the payment-lines module (which lists receivable /
payable accounts).

### Tests
- Distribution set on the payment lands on the counterpart line and nowhere else.
- Payment without distribution leaves all journal items clean.

@OCA/accounting-maintainers


rjaraspearhead:
ping @celm1990

celm1990 (approved):
LGTM.

@pedrobaeza @victoralmau, could you please review this? Should this feature be merged here or in the account-payment repository?

pedrobaeza:
The name should be IMO `account_payment_analytic`, for following the top-down rationale. It can be here, being analytic related.

rjaraspearhead:
> The name should be IMO `account_payment_analytic`, for following the top-down rationale. It can be here, being analytic related.

Rename done

celm1990:
In this PR, https://github.com/OCA/account-payment/pull/969, a new module with a similar feature is proposed. If this name doesn't conflict with another module, it's OK for me.

PR #964, opened 2026-08-19
[IMP] account_move_update_analytic: Update multiple move lines

hhgabelgaard:
via actions menu


Shide:
Can you add a test for this workflow?

PR #967, opened 2026-09-02
[UPD] Update oca-addons-repo-template to v1.43

weinni2000:
## Summary
- Bumps the OCA repo template (`.copier-answers.yml` `_commit`) from v1.39 to v1.43.
- Extracted as a standalone template-only update from the `account_analytic_distribution_manual` 19.0 migration in weinni2000/account-analytic#1, as suggested in review, since these files are copier-managed and unrelated to the module migration itself.

## Changes
- `.copier-answers.yml`: bump `_commit`, drop `ci: GitHub`, add `postgres_image: ''`
- `.github/workflows/pre-commit.yml`: bump Python to 3.12
- `.github/workflows/test.yml`: drop `OCA_ENABLE_CHECKLOG_ODOO` env var
- `.pre-commit-config.yaml`: pin `python3.12`, add "obsolete dotfiles" hook
- `checklog-odoo.cfg`: removed (no longer used)

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,8 +1,7 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.39
+_commit: v1.43
 _src_path: git+https://github.com/OCA/oca-addons-repo-template
 additional_ruff_rules: []
-ci: GitHub
 convert_readme_fragments_to_markdown: true
 enable_checklog_odoo: true
 generate_requirements_txt: true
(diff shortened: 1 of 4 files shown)


checklog-odoo.cfg
```diff
@@ -1,4 +0,0 @@
-[checklog-odoo]
```
pedrobaeza, line 1: This file is not correct that it's removed.

MiquelRForgeFlow (approved):

